PR #44, opened 2015-04-09
Ignore unsupported attributes that are supported by SQS

jcogilvie:
I'm sure there's a more idiomatic scala way to do what I'm doing, and it probably deserves some kind of flag for whether to enable it or not, but basically I created a list of all the unsupported-by-elasticmq-but-supported-by-SQS queue attributes, and before rejecting an update for an unsupported attribute we check if the attribute is valid in SQS, and then log+ignore it.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -10,4 +10,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.11.2")
 
 addSbtPlugin("com.gu" % "sbt-teamcity-test-reporting-plugin" % "1.5")
 
+addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "2.5.0")
+
 scalacOptions in ThisBuild ++= Seq("-unchecked", "-deprecation")
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -10,6 +10,4 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.11.2")
 
 addSbtPlugin("com.gu" % "sbt-teamcity-test-reporting-plugin" % "1.5")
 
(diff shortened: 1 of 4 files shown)


rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/QueueAttributesDirectives.scala
```diff
@@ -95,7 +105,14 @@ trait QueueAttributesDirectives { this: ElasticMQDirectives with AttributesModul
               case ReceiveMessageWaitTimeSecondsAttribute => {
                 queueActor ? UpdateQueueReceiveMessageWait(Duration.standardSeconds(attributeValue.toLong))
               }
-              case _ => Future(throw new SQSException("InvalidAttributeName"))
+              case _ => { // I'm sure there's a more idiomatic scala way to do this.
```
adamw: you could do:
````
case a if UAN.AAN.contains(a) => ...
case _ => Future(throw ...)
````

File: project/plugins.sbt [hunk 2]
```diff
@@ -10,4 +10,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.11.2")
 
 addSbtPlugin("com.gu" % "sbt-teamcity-test-reporting-plugin" % "1.5")
 
-scalacOptions in ThisBuild ++= Seq("-unchecked", "-deprecation")
\ No newline at end of file
+addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "2.5.0")
```
adamw: Local-environment-specific plugins should go to `~/.sbt/0.13/local.sbt`

adamw:
Thanks a lot! Just two minor remarks :)

jcogilvie:
There you go!

adamw:
Thanks!

Issue #58, opened 2015-09-18
Queue names are not limited to 80 characters


conniec:
Queue names are limited to 80 characters. Alphanumeric characters plus hyphens (-) and underscores (_) are allowed.

There should be some more validation on the SQS queue name when creating a queue. 

https://aws.amazon.com/sqs/faqs/

conniec:
hey @adamw will this be released in the next version? thanks!

adamw:
Released 0.8.12 :)

conniec:
awesome thanks! :)

PR #59, opened 2015-09-18
Queue with name longer than 80 chars should not be created.

conniec:
https://github.com/adamw/elasticmq/issues/58

### DIFF
--- a/core/src/test/scala/org/elasticmq/actor/QueueManagerActorTest.scala
+++ b/core/src/test/scala/org/elasticmq/actor/QueueManagerActorTest.scala
@@ -77,8 +77,6 @@ class QueueManagerActorTest extends ActorTest with QueueManagerForEachTest with 
     for {
       _ <- queueManagerActor ? CreateQueue(q1)
 
-      // When & then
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/elasticmq/actor/QueueManagerActorTest.scala
```diff
@@ -70,6 +70,20 @@ class QueueManagerActorTest extends ActorTest with QueueManagerForEachTest with
     }
   }
 
+  waitTest("trying to create a queue with long name should throw an exception") {
+    // Given
+    val q1 = createQueueData("q111111111111111111111111111111111111111111111111111111111111111111111111111111111", MillisVisibilityTimeout(1L))
+
+    for {
+      _ <- queueManagerActor ? CreateQueue(q1)
```
adamw: I suppose trying to create the queue once should be enough?

conniec: hmm yea I just took this from the other tests.

adamw:
Thanks!

PR #74, opened 2016-06-08
Updated readme file (link for akka-http)

slouc-bytecode:
Fixed link for akka-http (used to point to general akka.io)


adamw:
Thanks :)

PR #75, opened 2016-07-15
Update README with boto3 interface

fhennecker:
The boto3 interface has changed and there is barely any documentation anywhere on how to connect to a local elasticmq, here is a solution!


adamw:
Thanks! :)

Issue #88, opened 2016-12-05
Support for dynamic port


pjungermann:
This issue is related to #87 

Instead of using the default port, it would be great to support a dynamic port. I tried that by using port 0 and then retrieve the actual port from the started `SQSRestServer` instance, but due to #87, the queue addresses have been wrong and due to the fact that the `NodeAddress` is just some kind of string builder, using port 0 at the `serverAddress` just resulted in port 0 being part of the string.

A workaround here is to find a free socket yourself by i.e. using `new ServerSocker(0).getLocalPort`. For development, that was fine (so far), but there are concurrency issues as other process might bind to the port you have found to be free. Therefore, it was failing on our CI server. But exactly there, we would need that dynamic port feature to allow multiple test runs using ElasicMQ, e.g. feature branches or different projects.

nebnagrom:
I found it unclear how to get the port from the started server, for anyone else who tries to do this you'll need to wait for the server to start with something like this (Java):
`startedServer.waitUntilStarted().localAddress().getPort()`

adamw:
@nebnagrom maybe you could create a PR and add this info to the readme?

Issue #87, opened 2016-12-05
Changed port might cause wrong queue address


pjungermann:
When changing just the port `SQSRestServerBuilder`, the `serverAddress` will not get adjusted. Therefore, the resulting queue address is wrong. I used it for tests for an AmazonSQS client and got "queue not found" back. I also tested the address manual, and it was wrong - as it still used the default port.

`NodeAddress` defines the protocol, host, port and context to build a static address string. Still, this string should be in sync with the actual server instance.

This issue is related to #88

adamw:
Ah, I hoped that this would be more clear from the README, but when changing the port you need to change two things current: `SQSRestServerBuilder.withPort(myPort).withServerAddress(NodeAddress(port = myPort))`.

The reason is that the first specifies the port for binding, and the second the port returned in XML responses. These might be different: e.g. in a docker environment, you might bind to an interface/port, but to the outside world, the hostname/port might be different.

Do you have any suggestions on how to improve the API and keep the above functionality?

pjungermann:
I see, I didn't think about the docker use case. I was mainly focused on getting a dynamic port working as described in issue #88. 

I guess my PR is not working for the dockers use case, but was making the rest possible.

Will have a look tomorrow on how I can incorporate that. Maybe you can check it already. Am not much experienced with Scala, but should make it work I guess.

pjungermann:
PS: in my PR I also provided a new method `withDynamicPort`. It might be easy as using an additional dynamic port flag or just a check for port zero or another method asking for synchronized ports, etc.

adamw:
In 0.12.0, when using the Java/Scala API, changing the bind interface/port will also update the node address. Only when a custom node address is specified, bing interface/port won't affect it.

pjungermann:
Sounds great, thanks!

PR #89, opened 2016-12-05
dynamic port support (issue #88) + consistent port (issue #87)

pjungermann:
dynamic port support (issue #88) + consistent port (issue #87)
- new "withDynamicPort" method using port zero
- made env.serverAddress mutable in order to assign the bound address / port to it at the onSuccess of the app start


adamw:
Release in 0.12.0

PR #91, opened 2017-01-09
added dead letters queue support

mkorolyov:
Added dead letters queue support.
According to AWS console it is possible to configure dead letters queue for a dead letters queue, same we can do in elasticmq too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ queues {
         	deadLettersQueue2 {...}
     	}
     }
+    queue2 { }
 }
 ````
 
--- a/README.md
+++ b/README.md
@@ -122,19 +122,14 @@ queues {
         delay = 5 seconds
         receiveMessageWait = 0 seconds
         deadLettersQueue1 {
-            maxReceiveCount = 3 // requiredfrom 1 to 1000
-            // below all optional
-        	defaultVisibilityTimeout = 10 seconds
-        	delay = 5 seconds
-        	receiveMessageWait = 0 seconds
-        	deadLettersQueue2 {...}
+            maxReceiveCount = 3 // from 1 to 1000
     	}
     }
     queue2 { }
 }
 ````
 
-All attributes are optional except `maxReceiveCount` for dead letters queue, and if not specified will use the default SQS values.
+All attributes are optional except `maxReceiveCount` for dead letters queue.
 
 Starting an embedded ElasticMQ server with an SQS interface
 -----------------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -116,14 +116,24 @@ The queues are specified in a custom configuration file. For example, create a `
 ````
 include classpath("application.conf")
 
-queues {
-    queue1 {
+queues = [
+    {
+    	name = queue1
         defaultVisibilityTimeout = 10 seconds
         delay = 5 seconds
+        maxReceiveCount = 3
         receiveMessageWait = 0 seconds
-    }
-    queue2 { }
-}
+        deadLettersQueue = {
+        	name = deadletters1
+        	defaultVisibilityTimeout = 10 seconds
+        	delay = 5 seconds
+        	maxReceiveCount = 3
+        	receiveMessageWait = 0 seconds
+        	deadLettersQueue = {...}
+    	}
+    },
+    {...}
+]
 ````
 
 All attributes are optional, and if not specified will use the default SQS values.
--- a/README.md
+++ b/README.md
@@ -116,27 +116,24 @@ The queues are specified in a custom configuration file. For example, create a `
 ````
 include classpath("application.conf")
 
-queues = [
-    {
-    	name = queue1
+queues {
+    queue1 {
         defaultVisibilityTimeout = 10 seconds
         delay = 5 seconds
-        maxReceiveCount = 3
         receiveMessageWait = 0 seconds
-        deadLettersQueue = {
-        	name = deadletters1
+        deadLettersQueue1 {
+            maxReceiveCount = 3 // requiredfrom 1 to 1000
+            // below all optional
         	defaultVisibilityTimeout = 10 seconds
         	delay = 5 seconds
-        	maxReceiveCount = 3
         	receiveMessageWait = 0 seconds
-        	deadLettersQueue = {...}
+        	deadLettersQueue2 {...}
     	}
-    },
-    {...}
-]
+    }
+}
 ````
 
-All attributes are optional, and if not specified will use the default SQS values.
+All attributes are optional except `maxReceiveCount` for dead letters queue, and if not specified will use the default SQS values.
 
 Starting an embedded ElasticMQ server with an SQS interface
 -----------------------------------------------------------
--- a/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
@@ -104,7 +104,7 @@ trait QueueActorMessageOps extends Logging {
         None
       } else if (messagesById.contains(id.id)) {
         // Putting the msg to dead letters queue if exists
-        if (internalMessage.receiveCount >= queueData.maxReceiveCount) {
+        if (queueData.deadLettersQueue.flatMap(_.maxReceiveCount).exists(_ <= internalMessage.receiveCount)) {
           logger.info(s"send message $internalMessage to dead letters actor $deadLettersActorRef")
           deadLettersActorRef
             .map(_ ! SendMessage(internalMessage.toNewMessageData))
--- a/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
@@ -106,8 +106,7 @@ trait QueueActorMessageOps extends Logging {
         // Putting the msg to dead letters queue if exists
         if (queueData.deadLettersQueue.flatMap(_.maxReceiveCount).exists(_ <= internalMessage.receiveCount)) {
           logger.info(s"send message $internalMessage to dead letters actor $deadLettersActorRef")
-          deadLettersActorRef
-            .map(_ ! SendMessage(internalMessage.toNewMessageData))
+          deadLettersActorRef.foreach(_ ! SendMessage(internalMessage.toNewMessageData))
 
           // delete only if normal queue or if dead letter queue has its own dead letters queue
           if (!queueData.isDeadLettersQueue || (queueData.isDeadLettersQueue && deadLettersActorRef.isDefined)) {
--- a/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
@@ -103,35 +103,42 @@ trait QueueActorMessageOps extends Logging {
         messageQueue += internalMessage
         None
       } else if (messagesById.contains(id.id)) {
-        // Putting the msg to dead letters queue if exists
-        if (queueData.deadLettersQueue.flatMap(_.maxReceiveCount).exists(_ <= internalMessage.receiveCount)) {
-          logger.info(s"send message $internalMessage to dead letters actor $deadLettersActorRef")
-          deadLettersActorRef.foreach(_ ! SendMessage(internalMessage.toNewMessageData))
-
-          // delete only if normal queue or if dead letter queue has its own dead letters queue
-          if (!queueData.isDeadLettersQueue || (queueData.isDeadLettersQueue && deadLettersActorRef.isDefined)) {
-            internalMessage.deliveryReceipt.foreach(dr => deleteMessage(DeliveryReceipt(dr)))
-          }
-
-          return None
-        }
-
-        // Putting the msg again into the queue, with a new next delivery
-        internalMessage.deliveryReceipt = Some(DeliveryReceipt.generate(id).receipt)
-        internalMessage.nextDelivery = newNextDelivery.millis
-
-        internalMessage.receiveCount += 1
-        internalMessage.firstReceive = OnDateTimeReceived(new DateTime(deliveryTime))
-
-        messageQueue += internalMessage
-
-        logger.debug(s"${queueData.name}: Receiving message $id")
-
-        Some(internalMessage.toMessageData)
+        processInternalMessage(deliveryTime, newNextDelivery, internalMessage)
       } else {
         // Deleted msg - trying again
         receiveMessage(deliveryTime, newNextDelivery)
       }
+    }
+  }
+
+  private def processInternalMessage(
+    deliveryTime: Long, newNextDelivery: MillisNextDelivery, internalMessage: InternalMessage) = {
+    // Putting the msg to dead letters queue if exists
+    if (queueData.deadLettersQueue.flatMap(_.maxReceiveCount).exists(_ <= internalMessage.receiveCount)) {
+      logger.info(s"send message $internalMessage to dead letters actor $deadLettersActorRef")
+      deadLettersActorRef.foreach(_ ! SendMessage(internalMessage.toNewMessageData))
+
+      // delete only if normal queue or if dead letter queue has its own dead letters queue
+      if (!queueData.isDeadLettersQueue || (queueData.isDeadLettersQueue && deadLettersActorRef.isDefined)) {
+        internalMessage.deliveryReceipt.foreach(dr => deleteMessage(DeliveryReceipt(dr)))
+      }
+
+      None
+
+    } else {
+
+      // Putting the msg again into the queue, with a new next delivery
+      internalMessage.deliveryReceipt = Some(DeliveryReceipt.generate(MessageId(internalMessage.id)).receipt)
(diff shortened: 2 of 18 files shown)


core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
```diff
@@ -103,6 +103,15 @@ trait QueueActorMessageOps extends Logging {
         messageQueue += internalMessage
         None
       } else if (messagesById.contains(id.id)) {
+        // Putting the msg to dead letters queue if exists or delete.
+        if (internalMessage.receiveCount >= queueData.maxReceiveCount) {
+          logger.info(s"send message $internalMessage to dead letters actor $deadLettersActorRef")
+          deadLettersActorRef
+            .map(_ ! SendMessage(internalMessage.toNewMessageData))
+            .getOrElse(internalMessage.deliveryReceipt.foreach(dr => deleteMessage(DeliveryReceipt(dr))))
+          return None
```
adamw: [minor] maybe instead of the `return`, this could be an if-else. That way the `if` is an expression whose result is returned. Also this method is quite long - maybe a method could be extracted, e.g. `receiveMessage(id)`?

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
```diff
@@ -130,15 +130,15 @@ class QueueActorMsgOpsTest extends ActorTest with QueueManagerForEachTest with D
...
       // Then
-      lookupBeforeReceiving.flatMap(_.deliveryReceipt) should be (None)
+      lookupBeforeReceiving.flatMap(_.deliveryReceipt) should be(None)
 
       val receivedReceipt = received.flatMap(_.deliveryReceipt)
       val lookedUpReceipt = lookupAfterReceiving.flatMap(_.deliveryReceipt)
 
-      receivedReceipt.size  should be > (0)
-      lookedUpReceipt should be ('defined)
+      receivedReceipt.size should be > (0)
```
adamw: [minor] there's quite a lot of changes from reformatting which make it harder to spot the "real" changes

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
```diff
@@ -147,7 +147,7 @@ class QueueActorMsgOpsTest extends ActorTest with QueueManagerForEachTest with D
     val q1 = createQueueData("q1", MillisVisibilityTimeout(1L))
 
     for {
-      Right(queueActor) <- queueManagerActor ? CreateQueue(q1)
+      Right(queueActor) <- queueManagerActor ? CreateQueue(q1.copy(maxReceiveCount = 2))
```
adamw: is this change needed for the test to pass?

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
```diff
@@ -237,13 +238,13 @@ class QueueActorMsgOpsTest extends ActorTest with QueueManagerForEachTest with D
...
       // Then
-      lookupResult should be (None)
+      lookupResult should be(None)
     }
   }
 
   waitTest("msg statistics should be updated") {
     // Given
-    val q1 = createQueueData("q1", MillisVisibilityTimeout(1L))
+    val q1 = createQueueData("q1", MillisVisibilityTimeout(1L)).copy(maxReceiveCount = 2)
```
adamw: same here - is this needed?

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
```diff
@@ -298,7 +299,9 @@ class AmazonJavaSdkTestSuite extends FunSuite with Matchers with BeforeAndAfter
   test("should block message for the visibility timeout duration") {
     // Given
     val queueUrl = client.createQueue(new CreateQueueRequest("testQueue1")
-      .withAttributes(Map(defaultVisibilityTimeoutAttribute -> "1"))).getQueueUrl
+      .withAttributes(
+        Map(defaultVisibilityTimeoutAttribute -> "1", defaultMaxReceiveCountAttribute -> "3")))
```
adamw: Does this need to be specified in every test? Shouldn't an undefined value work, meaning no DLQ support?

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/CreateQueueDirectives.scala
```diff
@@ -13,6 +13,7 @@ import CreateQueueDirectives._
 
 trait CreateQueueDirectives { this: ElasticMQDirectives with QueueURLModule with AttributesModule with SQSLimitsModule =>
 
+  // DeadLettersQueue currently not supported via http call due to current API limitations. Consider moving to JSON?
```
adamw: Aren't DLQ's configured via attributes passed during queue creation? The docs (http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html) mention `RedrivePolicy`.  Where's the JSON?

server/src/main/scala/org/elasticmq/server/ElasticMQServerConfig.scala
```diff
@@ -57,20 +57,37 @@ class ElasticMQServerConfig(config: Config) {
   case class CreateQueue(name: String,
     defaultVisibilityTimeoutSeconds: Option[Long],
     delaySeconds: Option[Long],
-    receiveMessageWaitSeconds: Option[Long])
+    receiveMessageWaitSeconds: Option[Long],
+    deadLettersQueue: Option[CreateQueue] = None,
+    maxReceiveCount: Int = 1)
```
adamw: Aren't DLQ's optional? Then this should be optional too, no?

README.md
```diff
@@ -116,14 +116,24 @@ The queues are specified in a custom configuration file. For example, create a `
 ````
 include classpath("application.conf")
 
-queues {
-    queue1 {
+queues = [
```
adamw: If possible, I would prefer not to change the format of the `queues` config, so that people already using this won't have to migrate.

Some note in the readme about DLQ support and how it works would be great as well.

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
mkorolyov: sorry, my IDE reformatted code by itself, will try to revert such changes.

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
mkorolyov: yes, bc now number if receives is count and msg will not be available if we reach limit.

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
mkorolyov: yes, same

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
mkorolyov: this param added only to small part of tests, i added it int he way like `defaultVisibilityTimeoutAttribute` was done.
if needed we can introduce some DSL to make code more readable.

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/CreateQueueDirectives.scala
mkorolyov: didn't get it, RedrivePolicy expects a string with json like `"{\"maxReceiveCount\":\"5\", \"deadLetterTargetArn\":\"dead_letters_queue1\"}"` ?

core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
```diff
@@ -103,6 +103,20 @@ trait QueueActorMessageOps extends Logging {
         messageQueue += internalMessage
         None
       } else if (messagesById.contains(id.id)) {
+        // Putting the msg to dead letters queue if exists
+        if (internalMessage.receiveCount >= queueData.maxReceiveCount) {
+          logger.info(s"send message $internalMessage to dead letters actor $deadLettersActorRef")
+          deadLettersActorRef
+            .map(_ ! SendMessage(internalMessage.toNewMessageData))
```
sbinq: `foreach` is more idiomatic for side effects btw

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
adamw: but shouldn't there be no limit by default? As far as I understand, that's how SQS behaves

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/CreateQueueDirectives.scala
adamw: Ah, true! So the JSON is the value of the `RedrivePolicy` attribute. Then we should probably parse that :)

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
mkorolyov: seems there is inconsistent in sqs docs:
Here it says that maxReceiveCount is not required - http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues-redrivepolicy.html
While here it should be set from 1 to 1000: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues-redrivepolicy.html

core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
mkorolyov: yeah, sure, didn't change it in last fix commit, will update

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
adamw: I think you are looking at the docs of AWS CloudFormation, not SQS - maybe there's a typo in one of the CloudFormation docs. In the SQS docs (e.g. http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html), the `RedrivePolicy` is specified as *optional* (which it must be, to maintain backwards-compatibility).

This is also visible in the web interface:
http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html

core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
mkorolyov: agree, fixed

core/src/test/scala/org/elasticmq/actor/QueueActorMsgOpsTest.scala
mkorolyov: agree, thanks, fixed

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/CreateQueueDirectives.scala
mkorolyov: done

server/src/main/scala/org/elasticmq/server/ElasticMQServerConfig.scala
mkorolyov: agree, fixed

README.md
mkorolyov: done

core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
mkorolyov: done

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
```diff
@@ -894,6 +895,26 @@ class AmazonJavaSdkTestSuite extends FunSuite with Matchers with BeforeAndAfter
...
 
+  test("should create queue with redrive policy.") {
+    // When
+    val queueUrl = client.createQueue(new CreateQueueRequest("testQueue1")
+      .withAttributes(Map(
+        defaultVisibilityTimeoutAttribute -> "1",
+        redrivePolicyAttribute ->
+          """
+            |{
+            |  "queueName":"dlq1",
```
adamw: Keeping compatibility with SQS, shouldn't this be called `deadLetterTargetArn`? Also, the SQS docs (I'm looking here: http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html), don't mention the ability to specify e.g. the `defaultVisibilityTimeout`.

mkorolyov: i noticed that some other params in create queue request also has different name than in real sqs, so i decided to use more readable name, or i am wrong?
defaultVisibilityTimeout not mentioned in docs, but i was able to set it and all other params from AWS console.

adamw: they do? which ones? I think they should be the same ... if the tested code is supposed to work the same with real and mock SQS :)

can you set the `defaultVisibilityTimeout` when defining a dead letter queue for an existing queue? Or only when creating a queue?

mkorolyov:
@adamw 
> SQS requires the dead letters queue to be already created
didn't get it.
in AWS console there is a possibility to add dead letters queue to normal queue and you don't need to create dead letters queue manually before adding, it is created when you need automatically.

adamw:
I'm not that familiar with the console, I'm rather following the API docs, which ElasticMQ is supposed to follow (after all, we are implementing a mock SQS API implementation, not a mock SQS console :) )

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
mkorolyov: sorry, yes, all params has correct names according docs, will update new ones.

about `defaultVisibilityTimeout` both cases are valid. it can be updated at any time.

adamw: But if the dead letter queue has to be created before setting it as a DLQ, then why would you pass the visibility timeout here?

Anyway, thanks a lot for the changes! :)

mkorolyov:
@adamw checked one more time, yes, you are 100% right, dead letters queues should be created by hands before using.

So i need to update http handler a little bit not to create dead letters queues and return err if specified queue in RedrivePolicy is not created.

Will fix in a few days.
Thanks!

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
mkorolyov: yes, i was sure that dead letters queue should not be precreated, as it is wrong, i will remove all excess params from `RedrivePolicy` json.

mkorolyov:
@adamw 
Hey, had some time todays morning, updated PR according yesterday discussion.

mkorolyov:
@adamw when are you planning to make next release?

adamw:
@mkorolyov just release 0.13.0. I made a minor change to the config format, hope you don't mind :)

adamw:
Thanks a lot for your contribution!

mkorolyov:
thanks for fast feedbacks! 
BRs

joelrfcosta:
I'm working with version 0.13.2 `=== ElasticMQ server (0.13.2) started in 18185 ms ===` but I'm getting this message on console:
```
Ignored attribute "RedrivePolicy" (supported by SQS but not ElasticMQ)
```
Isn't `RedrivePolicy` functionality available since 0.13.0 @adamw?

adamw:
@joelrfcosta Ah, it's only supported in create queue right now. Can you create a bug for that?

PR #95, opened 2017-02-07
md5Digest(): explicitly provide getBytes() encoding

### LINKED ISSUE #13: MD5 Mismatch for some SQS messages
Firstly, thank you for making this excellent tool, it has been very useful.

I have found that some SQS messages when read through the Amazon SDK result in errors like:
MD5 hash mismatch for message id d97a1912-c5a3-4f41-8aa5-e6e37a87474f
   at Amazon.SQS.Util.AmazonSQSUtil.ValidateMD5(String message, String messageId, String md5FromService)
   at Amazon.SQS.Util.AmazonSQSUtil.ValidateMD5(Message message)
   at Amazon.SQS.Util.AmazonSQSUtil.ValidateReceiveMessage(ReceiveMessageResponse response)
   at Amazon.SQS.AmazonSQSClient.ReceiveMessage(ReceiveMessageRequest request)

I have experimented with the text I am sending. The following test case will result in that error. (It is worth adding that this message does work on the real amazon SQS)
"{\"Body\":\"{\\"Event\\":\\"{\\\\"RequestId\\\\":\\\\"b6889680-c986-4f99-9dff-a29000a800b1\\\\"\\\\"EmailSubject\\\\":\\\\"Toolbox – reset password\\\\"}\\"}\"}"

However other more complex messages go through fine.
Thank you,
Ivan

kopchik:
If encoding not provided, String.getBytes() uses system encoding to convert string into array of bytes. If system encoding is not set or it is not unicode, then unicode symbols may be lost/altered (we experienced this in our docker container). As a result, md5 checksum will be wrong (I'm not sure if the message itself gets corrupted). This should fix #13.

### DIFF
--- a/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala
+++ b/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala
@@ -246,7 +246,7 @@ object MD5Util {
   def md5Digest(s: String) = {
     val md5 = MessageDigest.getInstance("MD5")
     md5.reset()
-    md5.update(s.getBytes)
+    md5.update(s.getBytes("UTF-8"))
     md5.digest().map(0xFF & _).map { "%02x".format(_) }.foldLeft(""){_ + _}
   }
 


adamw:
Release as 0.13.2 - thanks! :)

PR #96, opened 2017-02-13
Documentation for working with dynamic ports

nebnagrom:
As per the comment in https://github.com/adamw/elasticmq/issues/88#issuecomment-278941474 i've added some documentation about working with dynamic ports


adamw:
Thanks!

PR #98, opened 2017-03-07
Update README.md

drichard:
Make it more obvious that you need to manually specify the dead letters queue in the `queues` section.

(It took me ages to figure out that elasticmq does not create the dead letters queue for you automatically.)


adamw:
Thanks & sorry for the trouble :)

PR #99, opened 2017-03-07
fix typo in example configuration for deadletter queues in README file

akhilpanchal:
Couldn't figure out why the queues with deadletter queue config in custom.conf file weren't being created, after following the example in README. Turns out there was a typo.


adamw:
Thanks! :)

PR #104, opened 2017-04-27
Update to Akka 2.5.0 and latest Akka HTTP

markhatton:


adamw:
Thanks! Though I think the current release is binary-compatible with 2.5.0, or is that not the case and a release is needed?

markhatton:
Sorry I neglected to provide any context ...

Akka HTTP 10.0.2 isn't compatible with Akka 2.5.0.  I get the following exception when using embedded ElasticMQ 0.13.2 with Akka 2.5.0:
```
java.lang.NoClassDefFoundError: Fatal execution error, caused by akka/stream/Fusing$
```
which is fixed here: akka/akka-http@2e54e0fa4f4474547a910543d7e8daa1d551523b

So the bump to Akka 2.5.0 isn't strictly necessary, but perhaps good hygiene?

Thanks

adamw:
Released in 0.13.3

Issue #97, opened 2017-03-07
ReceiveMessage does not filter out MessageAttribute


zsiddique:
http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html
So the AWS API is a little vague on this but testing shows that when ReceiveMessage is called with no MessageAttributeName defined ReceiveMessage returns messages with no MessageAttribute.  But elaticmq will return MessageAttribute regardless what is asked in MessageAttributeName.
This caused a problem for us as we are using elasticmq to test our usage of SQS and we had a bug in which we neglected to request an MessageAttribute but would work in testing but failed once we got to integration testing with real SQS.
Example using boto.  First with Elasticmq
```python
In [6]: c.send_message(MessageBody="ABody", MessageAttributes={'version': {  'StringValue': '1', 'DataType': 'String'}})
Out[6]:
{u'MD5OfMessageAttributes': '75cba0c5eb8ce78d6a41a35f99bfd7a8',
 u'MD5OfMessageBody': '3aff89437ced3559d5e62a20eeb3ea52',
 u'MessageId': '18791ce9-b66e-4ed5-ad31-e93064fcd673',
 'ResponseMetadata': {'HTTPHeaders': {'content-length': '579',
   'content-type': 'text/plain; charset=UTF-8',
   'date': 'Tue, 07 Mar 2017 00:30:02 GMT',
   'server': 'akka-http/2.3.15'},
  'HTTPStatusCode': 200,
  'RequestId': '00000000-0000-0000-0000-000000000000',
  'RetryAttempts': 0}}

In [7]: m = c.receive_messages()

In [8]: m[0].message_attributes
Out[8]: {'version': {u'DataType': 'String', u'StringValue': '1'}}
```

With AWS SQS:
```python
In [5]: c.send_message(MessageBody="ABody", MessageAttributes={'version': {  'StringValue': '1', 'DataType': 'String'}})
Out[5]:
{u'MD5OfMessageAttributes': '75cba0c5eb8ce78d6a41a3xxxxx',
 u'MD5OfMessageBody': '3aff89437ced3559d5e62a20xxxxxx',
 u'MessageId': '283c0711-6abd-42a5-b4d4-6b95c8cef099',
 'ResponseMetadata': {'HTTPHeaders': {'connection': 'keep-alive',
   'content-length': '459',
   'content-type': 'text/xml',
   'date': 'Tue, 07 Mar 2017 01:27:43 GMT',
[...]

PR #106, opened 2017-05-10
 #97: Make receive-message match sqs behaviour

DeepSpawn:
This resolves issue #97 

SQS only includes message attributes if they are specified in the receive message request, while ElasticMQ was returning all attributes by default.

I have updated the ReceiveMessageDirective to respect the messageAttributeName parameter and to no longer include attributes unless specified.


adamw:
Released in 0.13.4. Thanks!

Issue #110, opened 2017-05-19
MaxReceiveCount should be a string representation of an int, not an int


AjayTripathy:
See: [http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html)

It's surprising and a little unintuitive, but AWS expects a string representation of MaxReceiveCount:

String redrivePolicy = "{\"maxReceiveCount\":\"5\", \"deadLetterTargetArn\":\"arn:aws:sqs:us-east-2:123456789012:MyDeadLetterQueue\"}";

If you try to send a string over to elasticmq, however, elasticmq will complain that it expects an int.

adamw:
Indeed that should be corrected - maybe you could attempt a PR to fix that? :)

AjayTripathy:
Sure. Just wanted to make sure I had this right before starting.

PR #111, opened 2017-05-25
change maxReceiveCount to string to match AWS spec

AjayTripathy:
See https://github.com/adamw/elasticmq/issues/110


adamw:
Will be released soon in 0.13.5. Thanks!

PR #112, opened 2017-06-22
Fix queue creation with dead letter policy

### LINKED ISSUE #109: Deadletter Queues Should Be Assigned By ARN, not name.
I was able to create a deadletterqueue like this using boto3:
```
    def GetOrCreateQueue(self, queue_name):
        """Gets or creates a queue and a corresponding deadletter queue"""
        deadletter_queue_name = queue_name + '_deadletter'
        dlq = self.sqs_resource.create_queue(QueueName=deadletter_queue_name)
        redrive_policy = {
            'deadLetterTargetArn': queue_name + '_deadletter',
            'maxReceiveCount': 1
        }
        return self.sqs_resource.create_queue(QueueName=queue_name, Attributes={'RedrivePolicy': json.dumps(redrive_policy)})
```

But this is actually a small deviation from the sqs API-- deadletter queues should be configured by ARN, not name. The code should read:    
```
def GetOrCreateQueue(self, queue_name):
        """Gets or creates a queue and a corresponding deadletter queue"""
        deadletter_queue_name = queue_name + '_deadletter'
        dlq = self.sqs_resource.create_queue(QueueName=deadletter_queue_name)
        redrive_policy = {
            'deadLetterTargetArn': dlq.attributes['QueueArn'],
            'maxReceiveCount': 1
        }
        return self.sqs_resource.create_queue(QueueName=queue_name, Attributes={'RedrivePolicy': json.dumps(redrive_policy)})
```

Am I missing something here? The code on the bottom throws a queue not found exception when trying to create the second queue.

zozfried:
This should fix #109

### DIFF
--- a/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
+++ b/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
@@ -952,7 +952,7 @@ class AmazonJavaSdkTestSuite extends FunSuite with Matchers with BeforeAndAfter 
         redrivePolicyAttribute ->
           """
             |{
-            |  "deadLetterTargetArn":"dlq1",
+            |  "deadLetterTargetArn":"arn:aws:sqs:elasticmq:000000000000:dlq1",
             |  "maxReceiveCount":"1"
             |}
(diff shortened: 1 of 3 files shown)


adamw:
Released as 0.13.6. Thanks!

ealter:
Thanks for this fix. This caused some confusion during our elasticmq upgrade since we were previously passing an int for maxReceiveCount. This review caused it to only accept strings. However, it's technically more correct now since SQS's sdk also expects strings.

PR #113, opened 2017-06-23
Fix for MD5OfMessageAttributes when MessageAttributes is empty. This closes #114

### LINKED ISSUE #114: MD5OfMessageAttributes should not exist when MessageAttributes are empty
Hey,
When response doesn't contains `MessageAttributes` , the `MD5OfMessageAttributes` should be not added to response. (to be compatible with AWS SQS API). Now there is md5 sum of empty string (`d41d8cd98f00b204e9800998ecf8427e `).  It's not correct because official SDKs not allowed for this. 

[PHP SDK Validation](https://github.com/aws/aws-sdk-php/blob/master/src/Sqs/SqsClient.php#L212)
and after [it's returning null instead of MD5 of empty string](https://github.com/aws/aws-sdk-php/blob/master/src/Sqs/SqsClient.php#L125)

I found this same behaviour in some other official SDKs too, additionally official SQS API doesn't returns this argument in this case

The problem is in this particular line https://github.com/adamw/elasticmq/blob/master/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/ReceiveMessageDirectives.scala#L91

SebastianCzoch:
Hey!
Thank you for this awesome tool!
This PR solves https://github.com/adamw/elasticmq/issues/114

This is my first contribution to Scala project (in fact, first to Java similar thingy too) so, good hints are honestly welcomed.


adamw:
Will be soon released as 0.13.7. Thanks! Hopefully that's a start to more Scala/Java contributions :)

Issue #117, opened 2017-07-28
Custom QueueUrlContext


mal:
Currently when creating a queue the URL is always of the form `http://<host>/queue/<name>`. Originally I thought I could configure the `queue` part via the `context-path` setting, but that does not appear to be the case. Would it be possible to allow this [hardcoded](https://github.com/adamw/elasticmq/blob/80bc3978f943752143b4cc009902e0a638fd8b45/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala#L228) value to be customised?

This is also a request to be able to hit the URL generated with the custom `QueueUrlContext` too.

adamw:
Sure, that should be a matter of a simple configuration option. However the addresses returned by SQS always contain queue (unless something changed?), hence the current behavior.

mal:
Hmm, the queues I'm using on SQS are made up of `https;//<sqs-endpoint>/<aws-accound-id>/<queue-name>`. These are all on queues that live in the default region `us-east-1` and has dual working endpoints `queue.amazonaws.com` (legacy?) and `us-east-1.sqs.amazonaws.com`. Maybe it's a region inconsistency?

At any rate being able to change it in ElasticMQ would make for easier mocking as app code wouldn't have to distinguish between the global account ID used else where and the `queue` account it needs to use when in a mocked environment.

adamw:
Ah, maybe something changed since I originally wrote that or it's indeed region-specific. Anyway, adding such a config option should be pretty straightforward.

gpedic:
I would like to expand on this issue, AWS will actually allow you to omit the account id completely so both `https://<sqs-endpoint>/<aws-accound-id>/<queue-name>` and `https://<sqs-endpoint>/<queue-name>` will work while elasticmq always requires one to define the `QueueUrlContext` i.e. `queue`.

My code relies on being able to omit the account id at the moment so it would be great if elasticmq could simply accept requests without `QueueUrlContext` like `http://localhost:9324/<queue-name>` or simply alternatively rewrite them to `http://localhost:9324/queue/<queue-name>`.

Being able to remove `QueueUrlContext` completely via config for example by being able to define it as empty string could also be a simple solution.

sashaaKr:
Hey there, 
any progress on this issue?

micossow:
Hi, it's fixed in [1.3.13](https://github.com/softwaremill/elasticmq/releases/tag/v1.3.12)
[...]

PR #122, opened 2017-12-07
Queue actions work also with account id in urls

sergiu-terman:
even though the created queue urls have the format `/queue/MyQueue`. Existing apps
that use aws sqs operate on queues that also have the format
`/123123123123/MyQueue`. The introduced patch apply the queue actions
for queues that have both formats.

Here is an example of two commands that achieve the same result
```
aws sqs send-message --queue-url localhost:9432/queue/MyQueue --message-body "example 1"
aws sqs send-message --queue-url localhost:9432/12312ACCOUNT/MyQueue --message-body "example 1"
```

[related issue](https://github.com/adamw/elasticmq/issues/117)


Opalo:
@sergiu-terman can you point me to the docs where I can find the URL format?

Opalo:
Since will no longer accept PR to `master` branch directly I'm closing it and will create an issue for it.

PR #126, opened 2018-02-08
Add coverage instrumentation and few simple tests

simong:
This is a pretty simple PR that pulls in scoverage so that coverage reports can be generated after the tests have run. It's compatible with https://coveralls.io in case there's a desire to hook that up too.

For now, you can generate the reports by executing `sbt clean coverage test coverageReport`. That will instrument all the src/main/scala code and generate coverage reports for each subproject. Although it's mostly only core that's relevant, each report can be found at:
 - core/target/scala-2.12/scoverage-report/index.html
 - common-test/target/scala-2.12/scoverage-report/index.html
 - rest/rest-sqs/target/scala-2.12/scoverage-report/index.html
 - server/target/scala-2.12/scoverage-report/index.html

It's also possible to aggregate all reports by executing `sbt coverageAggregate` which will write an HTML report to target/scala-2.12/scoverage-report/index.html.

For core and the server, I've set minimum coverage thresholds so that if the coverage dips below those thresholds, `sbt coverageReport` will complain.

I'm working on adding in FIFO queue support and wanted to make sure I didn't miss anything :)

As a bonus, I've added 2 tests that bump up the coverage a bit.

### DIFF
--- a/core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
+++ b/core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
@@ -5,7 +5,7 @@ import org.elasticmq.actor.queue.QueueActor
 import org.elasticmq.actor.reply._
 import org.elasticmq.msg.{CreateQueue, DeleteQueue, _}
 import org.elasticmq.util.{Logging, NowProvider}
-import org.elasticmq.{QueueAlreadyExists, QueueData, QueueDoesNotExist}
+import org.elasticmq.{QueueAlreadyExists, QueueData}
 
 import scala.reflect._
 
@@ -20,9 +20,6 @@ class QueueManagerActor(nowProvider: NowProvider) extends ReplyingActor with Log
       if (queues.contains(queueData.name)) {
         logger.debug(s"Cannot create queue, as it already exists: $queueData")
         Left(new QueueAlreadyExists(queueData.name))
-      } else if (!queueData.deadLettersQueue.forall(dlq => queues.contains(dlq.name))) {
-        logger.error(s"Cannot create queue, its dead letters queue doesn't exists: $queueData")
-        Left(new QueueDoesNotExist(queueData.deadLettersQueue.get.name))
       } else {
         logger.info(s"Creating queue $queueData")
         val actor = createQueueActor(nowProvider, queueData)
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1,6 +1,7 @@
 import sbt._
 import Keys._
 import sbtassembly.AssemblyKeys._
+import scoverage.ScoverageKeys._
 
 object BuildSettings {
   val buildSettings = Defaults.coreDefaultSettings ++ Seq (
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
```diff
@@ -20,9 +20,6 @@ class QueueManagerActor(nowProvider: NowProvider) extends ReplyingActor with Log
       if (queues.contains(queueData.name)) {
         logger.debug(s"Cannot create queue, as it already exists: $queueData")
         Left(new QueueAlreadyExists(queueData.name))
-      } else if (!queueData.deadLettersQueue.forall(dlq => queues.contains(dlq.name))) {
```
adamw: so this code wasn't used? I see there's a test for that case, so I assume there's another path which returns the queue does not exist error message?

project/Build.scala
```diff
@@ -99,7 +100,11 @@ object ElasticMQBuild extends Build {
   lazy val core: Project = Project(
     "elasticmq-core",
     file("core"),
-    settings = buildSettings ++ Seq(libraryDependencies ++= Seq(jodaTime, jodaConvert, akka2Actor, akka2Testkit) ++ common)
+    settings = buildSettings ++ Seq(
+      libraryDependencies ++= Seq(jodaTime, jodaConvert, akka2Actor, akka2Testkit) ++ common,
+      coverageMinimum := 94,
```
adamw: We can monitor code coverage but I would be a bit wary of introducing required minimums. It would be good to know that coverage decreased, but ultimately I would trust the developers that they know what they are doing :)

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
```diff
@@ -959,6 +967,21 @@ class AmazonJavaSdkTestSuite extends FunSuite with Matchers with BeforeAndAfter
...
+      client.createQueue(new CreateQueueRequest("q1").withAttributes(Map(redrivePolicyAttribute ->
+          """
+            |{
+            |  "deadLetterTargetArn":"arn:aws:sqs:elasticmq:000000000000:queueDoesNotExist",
+            |  "maxReceiveCount":"1"
+            |}
+          """.stripMargin
+      )))
+    }
+  }
```
adamw: Thanks for the tests, always a great addition! :)

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
simong: That's right, it's currently checked at https://github.com/adamw/elasticmq/blob/master/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/CreateQueueDirectives.scala#L40

adamw: Ah, that's a very good outcome of the coverage changes, thanks :)

project/Build.scala
simong: FWIW, this will only affect the `coverageReport` task, after it generated the reports, it will exit with a RunTimeException. As that's usually the last task in the chain, it shouldn't affect much. Tasks such as `sbt test` will still work as before. That said, I can see why it can be a bit scary, so I'll take it out. We can always add it back in later when we're a bit more comfortable with it.

adamw: Thanks for the explanation :) By the way - maybe you could add the information on how to generate the coverage reports (as in the PR description) to the readme?

simong: Good idea! Added that in :)

Issue #124, opened 2018-01-04
Redrive Policy Example 


harry9111985:
Hi adamw,

If you have any samples to create redrivepolicy links using dead queues that would be great . Currently , I am trying to create a redrivepolicy during CreateQueueRequest (Amazon Java SDK) using queue attributes. I do get a debug point on the RedrivePolicy.class in the elasticmq package . But elasticmq doesnt seem to create a link with the DeadLetterQueue. 

My java code : 

final Map<String, String> deadLetterQueueAttrs = new HashMap<>();
		deadLetterQueueAttrs.put("VisibilityTimeout", "600");
		final CreateQueueRequest createDeadLetterQueueRequest = new CreateQueueRequest()
			.withQueueName(queueName + "-error")
			.withAttributes(deadLetterQueueAttrs);

		final CreateQueueResult deadLetterQueueResult = amazonSQSAsyncClient.createQueue(createDeadLetterQueueRequest);
		final GetQueueAttributesResult deadLetterQueueArnResult = amazonSQSAsyncClient.getQueueAttributes(deadLetterQueueResult.getQueueUrl(), Arrays.asList("QueueArn"));

		final Map<String, String> queueAttributes = new HashMap<>();
		final StringBuilder redrivePolicyParamBuilder = new StringBuilder()
			.append("{\"deadLetterTargetArn\" : \"")
			.append(deadLetterQueueArnResult.getAttributes().get("QueueArn"))
			.append("\",")
			.append("\"maxReceiveCount\": \"1\"}");
		queueAttributes.put("RedrivePolicy", redrivePolicyParamBuilder.toString());
		queueAttributes.put("VisibilityTimeout","600");

		final CreateQueueRequest createQueueRequest = new CreateQueueRequest()
			.withQueueName(queueName)
			.withAttributes(queueAttributes);

		final CreateQueueResult createQueueResult = amazonSQSAsyncClient.createQueue(createQueueRequest);

No errors are thrown but the redrive policy link is not created,

Kindly help.

Regards,
Harish

adamw:
Hello,

here's a test which uses a redrive policy:

https://github.com/adamw/elasticmq/blob/master/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala#L944-L960

Maybe it will help?

Adam

> On 4 Jan 2018, at 07:10, harry9111985 <notifications@github.com> wrote:
> 
[...]

PR #131, opened 2018-03-07
#124 - Output RedrivePolicy attributes

simong:
Saw the chatter on the ticket and figured this would be an easy one to familiarize myself with the codebase a bit more 😊

### DIFF
--- a/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
+++ b/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
@@ -6,15 +6,16 @@ import org.scalatest.Matchers
 import org.scalatest._
 import com.amazonaws.auth.BasicAWSCredentials
 import com.amazonaws.services.sqs.{AmazonSQS, AmazonSQSClient}
-
 import scala.collection.JavaConversions._
 import scala.collection.JavaConverters._
 
 import com.amazonaws.services.sqs.model._
 import scala.util.control.Exception._
+
 import com.amazonaws.AmazonServiceException
 import org.elasticmq.util.Logging
 import org.elasticmq._
+import org.elasticmq.rest.sqs.model.RedrivePolicy
 
 class AmazonJavaSdkTestSuite extends FunSuite with Matchers with BeforeAndAfter with Logging {
   val visibilityTimeoutAttribute = "VisibilityTimeout"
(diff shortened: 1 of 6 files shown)


rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/model/RedrivePolicyJson.scala
```diff
@@ -0,0 +1,26 @@
...
+
+
+object RedrivePolicyJson extends DefaultJsonProtocol {
+
+	implicit val format: JsonFormat[RedrivePolicy] = new RootJsonFormat[RedrivePolicy] {
+		def read(json: JsValue): RedrivePolicy = {
+			json.asJsObject.getFields("deadLetterTargetArn", "maxReceiveCount") match {
+				case Seq(JsString(deadLetterTargetArn), JsString(maxReceiveCount)) =>
+					RedrivePolicy(deadLetterTargetArn.split(":").last, maxReceiveCount.toInt)
+				case Seq(JsString(deadLetterTargetArn), JsNumber(maxReceiveCount)) =>
```
simong: I added this as the docs technically specify a number (but then oddly use a string in the examples).

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/model/RedrivePolicyJson.scala
```diff
@@ -0,0 +1,26 @@
+/*
+ * All right reserved - Fronteer LTD
```
adamw: I'm not sure that applies when contributing to an open-source project? (maybe it doesn't, but I'm not very familiar with copyright law :) )

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
```diff
@@ -950,34 +951,34 @@ class AmazonJavaSdkTestSuite extends FunSuite with Matchers with BeforeAndAfter
...
     a[QueueDoesNotExistException] shouldBe thrownBy {
-      client.createQueue(new CreateQueueRequest("q1").withAttributes(Map(redrivePolicyAttribute ->
-          """
-            |{
-            |  "deadLetterTargetArn":"arn:aws:sqs:elasticmq:000000000000:queueDoesNotExist",
-            |  "maxReceiveCount":"1"
-            |}
-          """.stripMargin
+      client.createQueue(new CreateQueueRequest("q1").withAttributes(Map(
+        redrivePolicyAttribute -> RedrivePolicy("queueDoesNotExist", 1).toJson.toString()
```
adamw: Maybe a test that the `RedrivePolicy` converts to the expected json would be good to have as well? Since we kind of tested this here - giving the raw json to the request, and now  I think there's no place where the to-json conversion is checked

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/model/RedrivePolicyJson.scala
simong: Oh dear, I must have a global template for new files. I'll take this out, good catch!

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
simong: I've expanded the tests a bit in `RedrivePolicyJsonTest` for verifying (de)serializing of `RedrivePolicy` instances.

simong:
Thank you for the review, I pushed some follow-up.

adamw:
released in 0.13.9

PR #136, opened 2018-04-27
Add a simple Dockerfile

willejs:
Added a simple dockerfile so that people can easily build this into a container. 
As its using a file off S3, and not built in CI I have added checksumming too.

Ideally there would be some ci building the code and dockerfile and cutting releases, which im happy to do if i end up using this a lot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM java:8
+
+ARG VERSION=0.13.8
+ARG VERSION_SHASUM=41525fa141f36145e4bb54211bee0906a1901d1cfcbb0d86b8f4367b521e69e1
+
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! Maybe it would be more clear to save these files e.g. in a `docker` subdirectory instead of the root?

adamw:
@willejs yes looks good, thanks! :)

cajund:
Folks, are you or do you plan to generate an official container image and store it on Docker Hub? I noticed there are a number of other folks doing this, but I would rather use something directly from the source. It's an extra step, but it would be helpful. Thanks!

adamw:
@cajund at some point, definitely, the backlog just needs to get a bit shorter ;)

Issue #137, opened 2018-05-15
Messages always returned in reverse order, intentional?


owst:
When receiving multiple messages from elasticmq, the messages are _always_ returned in reverse-insertion order - is this a conscious design decision?

This behaviour is allowed by the [SQS FAQs](https://aws.amazon.com/sqs/faqs/) (see "Does Amazon SQS provide message ordering?"), as standard SQS queues only "provide a loose-FIFO capability that attempts to preserve the order of messages" and therefore elasticmq is _not_ breaking any "specification". But it seems somewhat strange to actively return the messages in reverse order, rather than making any attempt at preserving insertion order (which is trivial to do, see below).

# Example

Consider the following simple test script that relies on the [aws cli](https://aws.amazon.com/cli/) and [`jq`](https://stedolan.github.io/jq/) and assumes `elasticmq` is available at `http://localhost:9324`:

```bash
#!/usr/bin/env bash
set -euo pipefail

queue_url=$(aws --endpoint-url http://localhost:9324 sqs create-queue --queue-name test-queue | jq -r .QueueUrl)

function sqs() {
    aws --endpoint-url http://localhost:9324 sqs $@ --queue-url $queue_url
}

function send() {
    message_id=$(sqs send-message --message-body $1 | jq .MessageId)
    echo "Sent message $message_id with body $1"
}

function receive() {
    echo "Receiving at most $1 messages..."
    sqs receive-message --max-number-of-messages $1 | jq '.Messages | .[] | .Body'
}

send 1-1
send 1-2
send 1-3
send 1-4

receive 1
receive 1
receive 1
receive 1

send 2-1
send 2-2
send 2-3
send 2-4

receive 2
receive 2

send 4-1
send 4-2
send 4-3
send 4-4

receive 4

sqs delete-queue --queue-url $queue_url
```

Here, we send 4 messages three times, receiving one at a time, then two at a time and finally four at a time; an example output as per ca10395bb9e7f39f3a687174f2c63d9908943bcd is:
```
Sent message "5d8a80f5-a60a-4a90-94d1-de050437426c" with body 1-1
Sent message "2b63922d-9e41-41fd-b581-87101e6440f4" with body 1-2
Sent message "9824ae60-c02f-4b9f-a115-70181bf23c6d" with body 1-3
Sent message "6a9699d1-a658-4381-8159-3bc1a6992e30" with body 1-4
Receiving at most 1 messages...
"1-1"
Receiving at most 1 messages...
"1-2"
[...]

PR #138, opened 2018-05-16
#137 - Do not return multiple messages in reverse order

owst:


adamw:
Released im 0.13.10, thanks!

simong:
Hey @adamw, is it possible this didn't quite make it out to the maven repos? http://central.maven.org/maven2/org/elasticmq/elasticmq-rest-sqs_2.12/ doesn't seem to list it.

Thanks!

adamw:
@simong sorry, should be there shortly :)

PR #141, opened 2018-06-08
Explicitly wait for server unbind before actor system termination

### LINKED ISSUE #140: Potential Terminate/Unbind race condition in shutdown of SQSRestServer
I have been intermittent experiencing issues getting the error `java.lang.IllegalStateException, with message: IO Listener actor terminated unexpectedly for remote endpoint` when running with the `SQSRestServer` (examples at: https://github.com/akka/akka-http/issues/1837, https://github.com/akka/alpakka/issues/913 & https://github.com/akka/alpakka/issues/915).

It looks like a race condition between the actor system shutdown and the server unbinding where the server received the `Terminated` message before the `Unbind`. The cause of this seems to be the ordering of the shutdown in the `SQSRestServer` created in the `start()` method of `TheSQSRestServerBuilder`.

randomcoder:
By at least waiting for the unbind request to complete we try and ensure that the server has received the `Unbind` request before the actor system sends a `Terminated` message. Thus we should avoid the underlying race condition.

fixes #140

### DIFF
--- a/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala
+++ b/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala
@@ -6,7 +6,7 @@ import java.security.MessageDigest
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicReference
 
-import akka.actor.{ActorRef, ActorSystem, Props}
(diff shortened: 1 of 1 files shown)


randomcoder:
@adamw I was wondering when the 0.13.11 will be published please? Maven central/Sonatype still show only 0.13.10 at the moment.

adamw:
Sorry, sth is wrong with the release process. Should be fixed soon :)

PR #143, opened 2018-06-12
FIFO support

simong:
As this is getting rather large, I'm submitting this for review now. I believe this is now at a point where it support most of what SQS' FIFO queues do.

I would be very grateful for any review. I understand it's a rather large diff, but I hope that some of the refactoring that's in there, makes the code a bit easier to follow. Happy to run through it in more detail if you want!


adamw:
That's awesome work - thanks! There isn't really anything that I could pick on :) Are you planning to adding anything more or is this ready to merge?

adamw:
I'm assuming that's all, merging

adamw:
Released as 0.14.0, thanks again!

simong:
Thank you! I really enjoyed working on this feature and the project in general, will definitely try and keep up with the latest developments :)

adamw:
Looking forward to more PRs :)

PR #144, opened 2018-06-20
Remove println statement from ReceiveMessageDirectives

simong:
I've no idea how this slipped through 😔


adamw:
Same here! And I though I read the code quite carefully :)

Issue #121, opened 2017-11-08
Support a mirror queue for testing


nhoughto:
We are using ElasticMQ as local SQS substitute for testing, one problem we have is in a full integrate test we are testing both the producers and consuming components and asserting a message was created is racey as it is often consumed before the assertion runs.

Lots of ways to tackle this, one would be for ElasticMQ to mirror all messages received on a specific queue to a mirror queue that would only be consumed by the test/assertions themselves. Is such a thing possible / desirable? Doing it in ElasticMQ rather than the client code seems cleaner.

adamw:
That's not currently supported, but adding an "audit" queue to verify that all messages have been properly handled could make much sense. It shouldn't be very hard to implement, if you'd like to try :)

nhoughto:
Does the “delay” setting try to achieve the same goal? If I enable a delay on the queues does that give me a window to assert again the message being in the queue before it’s read?

I tried delay but it didn’t seem to do anything, messages were just read straight away.

adamw:
I think the original purpose of delay is different, to deliver messages after a certain period of time, but I suppose it should be possible to implement what you are after using this as well.

If I recall correctly, a global queue-delay should apply to all messages, unless they have a different delay explicitly specified. Is that not the behavior you are seeing?

nhoughto:
I've got it configured like this

```
queues {
  transactions {
    delay = 10 seconds
  }
}
```

And with DEBUG level i'm expecting to see a `Sent..` event, and no `Receive..` event until 10+seconds afterwards. But i see the `Receive..` immediately afterwards. I couldn't see anywhere to specify a global delay (which i would prefer), how do I do that?

nhoughto:
Either way, as your mentioned `delay` doesn't achieve what i want, it would just make the messages unavailable to both the consumer and the test assertion. The mirror / audit queue is a better solution, any tips for how to PR it in? My scala is weak.

adamw:
There's no way to create a global delay; and I think the way you configured the queue should work fine; maybe that's a separate bug ;)

Anyway, you are right that using the delay (even if it worked properly) wouldn't do the trick here, as the messages would be unavailable to both consumer.

I think the way to go here would be to add an `auditQueue`, similarly to how a `deadLettersQueue` is specified. This would require changes in the configuration, server, all the way up to the core actors. The Scala isn't very sophisticated, but it would be probably a couple of hours of work :)

nhoughto:
[...]

PR #145, opened 2018-07-23
Implement mirror queues support configured via copyTo/moveTo attributes

sbinq:
This should resolve issue #121


adamw:
Released in 0.14.2. Thanks!

Issue #132, opened 2018-03-14
Ignored attribute "RedrivePolicy"


carymrobbins:
I'm confused about Dead Letter Queue support. According to this issue https://github.com/adamw/elasticmq/issues/124#issuecomment-370620243 it seems it should work; however, when running elasticmq 0.13.8 (via sbt-sqs) I see this warning when setting the RedrivePolicy -

```
WARN  o.e.r.s.TheSQSRestServerBuilder$$anon$1 - Ignored attribute "RedrivePolicy" (supported by SQS but not ElasticMQ)
```

Might I just be doing something wrong?

jaxley:
Looks like the code is set to be part of 0.13.9, which is not yet in the repositories and is only tagged here, but not listed as a release.

Issue #69, opened 2016-03-31
Support for Redrive Policy/dead letter queues


f24-gklein:
Hi! 
First of all: Thanks for this amazing work!

It would be nice if elasticmq had support for sqs redrive policy and dead letter queues, e.g. in the following places:
- getQueueAttributes
- setQueueAttributes
- also add support for setting 'maxReceiveCount' and 'deadLetterTargetArn' via elasticmq queue configuration file. 
- Correct dead letter queue configuration should be implemented in the 'business logic'  accordingly

See aws doc at: http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/SQSDeadLetterQueue.html

Regards, Gunther

travishumphreys:
Seconded! Especially because I'm trying to testing a queue with redrive policy, and create queue will return QueueNameExists after queue is created because attributes don't match.

norcalli:
Thirded, so that we could match our production server in testing.

AjayTripathy:
Was this resolved in https://github.com/adamw/elasticmq/pull/91 , or is only static configuration via the conf file supported for now?

adamw:
@AjayTripathy DLQs are only supported when creating queues (either via the API or static config). Adding support for get/set attribute should be an easy change, though

seanf:
I think this can be closed now. DLQs can now be created via setAttribute too (#147).

It appears that a message won't be moved to the DLQ until a client tries to receive it again (after exceeding `maxReceiveCount`). I'm not sure if that's standard SQS behaviour.

Opalo:
Thanks @seanf - closing then :)

PR #147, opened 2018-08-02
Update DLQ to allow adding after queue creation and allow tagging queue

mhughesbv:
I've altered the DLQ behaviour to be more in line with the AWS SQS specification so that a dead letter queue can be added after a queue has been created.

Alongside this I've added support for tagging in a similar fashion to AWS SQS, allowing freeform tags in a String,String format.

This should address #69 and #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ queues {
         contentBasedDeduplication = false
         copyTo = "audit-queue-name"
         moveTo = "redirect-queue-name"
+        tags {
+            tag1 = "tagged1"
+            tag2 = "tagged2"
+        }
     }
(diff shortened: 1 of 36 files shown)


adamw:
Thanks a lot! There's a lot of changes :)
For the future, it's usually better to submit "code-improvement" and functionality PRs separately - it makes them easier to review, and easier to focus on the specific change being proposed.

As for this PR, it can stay as-is, though there's a `TODO` in `CreateQueueDirectives` - should it stay that way?

adamw:
Thanks again for the changes! & looking forward for more

Release as 0.14.4 (also as a Docker image)

Issue #151, opened 2018-08-30
Long Polling Requests Fail on FIFO Queues


chrislbs:
We're using elasticmq to mock the SQS service for our java application. We're using the Java JMS AWS SDK which uses long polling 'WaitTimeSeconds=20 ' to prefetch messages for processing. We noticed during our tests that the time from message sent -> message received that it would take upwards of 15 seconds (or sometimes not at all) for us to get an answer on FIFO queues but standard queues always seemed pretty snappy e.g. < 1 second. I'm not familiar enough with Scala or Akka to really tackle this issue and understand what's causing the problem but I wrote a test that at least simulates the behavior I'm seeing.

```scala
// THIS TEST FAILS
test("FIFO queues should respond during long polling") {
    // Given
    val createRequest = new CreateQueueRequest(s"testFifoQueue-long-poll.fifo")
      .addAttributesEntry("FifoQueue", "true")
      .addAttributesEntry("ContentBasedDeduplication", "false")

    val queueUrl = client.createQueue(createRequest).getQueueUrl

    val messageRequest = new SendMessageRequest(queueUrl, "Message 1")
      .withMessageGroupId("group1")
      .withMessageDeduplicationId(UUID.randomUUID().toString)

    val longPollRequest = new ReceiveMessageRequest(queueUrl).withWaitTimeSeconds(20);

    var t = new Thread() {
      override def run(): Unit = {
        Thread.sleep(3000L)
        client.sendMessage(messageRequest)
      }
    }
    t.start()

    // When
    var start = System.currentTimeMillis()
    var messages = client.receiveMessage(longPollRequest).getMessages
    var end = System.currentTimeMillis()

    // Then
    (end - start) should be >= 3000L
    (end - start) should be <= 6000L
    messages.size should be(1)

    // run against with same content and group id but different deduplication id
    messageRequest.withMessageDeduplicationId(UUID.randomUUID().toString)

    t = new Thread() {
      override def run(): Unit = {
        Thread.sleep(3000L)
        client.sendMessage(messageRequest)
      }
    }
    t.start()

    // When
    start = System.currentTimeMillis()
    messages = client.receiveMessage(longPollRequest).getMessages
    end = System.currentTimeMillis()

    // Then
    (end - start) should be >= 3000L
    (end - start) should be <= 6000L
    messages.size should be(1)
  }

// THIS TEST PASSES
test("Standard queues has no issue during long polling") {
    // Given

[...]

PR #155, opened 2018-10-11
151: Fixing long polling FIFO requests using a requestAttemptId

chrislbs:
The fifo request attempt cache is being populated by an empty list which
would cause subsequent attempts to retrieve messages to always return an
empty list as it would never go back to the queue. This change ensures
that if the cache returns an empty list an attempt to dequeue a message
will utilize the message queue itself rather than the cache.


chrislbs:
Address #151

adamw:
Release in 0.14.6, thanks! :)

Issue #166, opened 2019-02-21
Queue actions work also with account id in urls.


Opalo:
This issue is related to #122.

PR #168, opened 2019-02-21
Changed regex to match queue URL format.

Opalo:
Related to #122 and #166.


adamw:
How about a test? You could create a queue (using the sqs sdk), and then issue a request to a hand-craft URL using an account number

Opalo:
@adamw test is in progress ;) Could you please give more detailed description of the test?

adamw:
sth like here: https://github.com/softwaremill/elasticmq/blob/develop/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala#L140-L149, only not deleting the queue but checking that it exists using a hand-created URL

Opalo:
Cool! That's what I thought :)

PR #169, opened 2019-02-21
Refactor ActorTest to allow running single test separately

rwalerow:
This request allows running tests separately in IDE like IntelliJ idea


rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/directives/AnyParamDirectives.scala
```diff
@@ -23,5 +23,11 @@ trait AnyParamDirectives {
     }
   }
 
+  def region: Directive1[Option[String]] = {
```
adamw: how come is it part of this pr? :)

rwalerow: My bad, I removed this code

Issue #133, opened 2018-04-03
Queue config supported in embedded mode


krisso-b:
While starting embedded server we should be able to provide queue's configuration.
It's already possible to provide such parameter in standalone mode:
```
-Dconfig.file=custom.conf
```

More or less it should look like
```
SQSRestServerBuilder
  .withConfigFile("classpath:/com/company/custom.conf")
  .start();
```

Alternatively/additionally we may consider providing file's content:
```
String config = "queues { ... }"; // normally shouldn't be hardcoded

SQSRestServerBuilder
  .withConfig(config)
  .start();
```

TheNickBartlett:
Been looking for this feature for a while now. Would love to see it happen!

michalo2882:
`SQSRestServerBuilder` does not do any autoconfiguration based on config.
You should used `ElasticMQServer` as described in [README](https://github.com/softwaremill/elasticmq/blob/master/README.md#starting-an-embedded-elasticmq-server-with-an-sqs-interface).

PR #214, opened 2019-06-20
#133 add note about config in embedded mode

RichardBradley:
It was not clear to me that embedded mode does not load config. The current Readme suggests that it will, to my reading, which confused me for a while. I thought I had named my config file wrong and spent a while trying to debug the config code.


adamw:
Yes that's correct, thanks for the docs update!

PR #268, opened 2019-08-18
Preserve ordering in batch requests.

janjaali:
Sending a batch of messages to a queue may result in a not-ordered processing of the received messages which may cause not expected behaviours in the case of FIFO queues (see #267).

The `SendMessageBatchDirectives` attempts to retrieve the list of provided parameters via the `BatchRequestsModule` which uses internally a `mutable.Map` to store handled parameters. This result-map does not maintain the order of the entry parameter list (please see the provided test). This PR aims to modify the `BatchRequestsModule::subParametersMaps` method in such a way that it preserves the order by sorting the parameters accordingly to the discriminator.


adamw:
Released in 0.14.8, thanks!

PR #269, opened 2019-08-19
Fix ordering in batch requests greater than or equal 10.

janjaali:
Fixes the sorting introduced in #268  for batches with a size greater than or equal 10. 

The currently implemented sorting mechanism relays on a `String`-sort which may lead to incorrect sorting behaviour for numbers greater than or equal 10. This PR will fix the sorting by transforming the discriminator to a `Int` before the sorting applies. In the special case of AWS SQS we should be able to do this without to worry about `NumberFormatException`s, see [SendMessageBatch](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html
).

Please note that AWS SQS has currently no support for batches with a size greater than 10 by design.

### DIFF
--- a/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/BatchRequestsModule.scala
+++ b/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/BatchRequestsModule.scala
@@ -73,6 +73,6 @@ object BatchRequestsModule {
         }
     }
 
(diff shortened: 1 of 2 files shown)


janjaali:
Sorry @adamw that I noticed the error now, was too fast with my PR. I hope the comments and the test are sufficient enough now.

adamw:
No problem - should be fixed in 0.14.10

Issue #273, opened 2019-08-26
Cannot send message to FIFO queue with DelaySeconds=0 parameter


mclark1129:
When a SendMessage request is sent to a FIFO queue which contains the `DelaySeconds` parameter, the request fails with a 400 response.  According to the AWS docs you should not be able to provide `DelaySeconds` for a FIFO queue, so for the most part this behavior makes sense.  However, if you provide a value of 0 seconds then this request is not rejected by Amazon.  Elastic MQ should only throw an error if DelaySeconds exists and is any value other than 0.

HTTP Request Body:
```
Action=SendMessage&MessageBody=Test&MessageDeduplicationId=DedupeId&MessageGroupId=UniqueGroupId&DelaySeconds=0
```

```
<ErrorResponse xmlns="http://queue.amazonaws.com/doc/2012-11-05/">
      <Error>
        <Type>Sender</Type>
        <Code>InvalidParameterValue</Code>
        <Message>InvalidParameterValue; see the SQS docs.</Message>
        <Detail/>
      </Error>
      <RequestId>00000000-0000-0000-0000-000000000000</RequestId>
    </ErrorResponse>
```

if you remove the `DelaySeconds` parameter from the request, then the message is queued successfully.

adamw:
Released in 0.14.11

PR #274, opened 2019-08-26
Allow delay seconds of 0 on FIFO queues + expand error messages

simong:
Addresses #273

I took the liberty to also expand some of the existing error messaging (inline with what AWS returns of course) as it's sometimes confusing why a FIFO request might be failing.


adamw:
Released in 0.14.11. Thanks!

PR #275, opened 2019-08-26
More precise creation timestamp for internal messages

janjaali:
Internal messages are created with a creation-timestamp truncated to milliseconds (see [DateTime](https://github.com/JodaOrg/joda-time/blob/5f98d636174cc0c6b76c9f434ee17e2b36577414/src/main/java/org/joda/time/DateTime.java#L36)) and are compared with each other using this timestamp. Since the implementation of the FIFO queue uses a internally a `PriorityQueue` for queuing and this `PriorityQueue` does not guarantee the order of queued messages for messages with the same order (see [PriorityQueue](https://www.scala-lang.org/api/2.13.0/scala/collection/mutable/PriorityQueue.html)) it may happen that messages received in a batch may not be ordered corresponding to their order in the batch. 

Since the AWS SQS guarantees that messages within a single batch are enqueued in the order they are sent (see [SendMessageBatchRequest](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/AmazonSQS.html#sendMessageBatch-com.amazonaws.services.sqs.model.SendMessageBatchRequest-)) this PR attempts to mirror this guarantee in ElasticMQ by providing a more precised creation-timestamp called `PreciseDateTime` which relays upon the already used `DateTime` and an additional nano seconds precision.

Relates to #267.

### DIFF
--- a/core/src/main/scala/org/elasticmq/NewMessageData.scala
+++ b/core/src/main/scala/org/elasticmq/NewMessageData.scala
@@ -6,5 +6,6 @@ case class NewMessageData(
     messageAttributes: Map[String, MessageAttribute],
     nextDelivery: NextDelivery,
     messageGroupId: Option[String],
-    messageDeduplicationId: Option[String]
+    messageDeduplicationId: Option[String],
+    orderIndex: Int
 )
--- a/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
@@ -3,7 +3,6 @@ package org.elasticmq.actor.queue
 import java.util.UUID
 
 import org.elasticmq._
-import org.elasticmq.actor.queue.InternalMessage.PreciseDateTime
 import org.elasticmq.util.NowProvider
 import org.joda.time.DateTime
 
@@ -15,7 +14,8 @@ case class InternalMessage(
     var nextDelivery: Long,
     content: String,
     messageAttributes: Map[String, MessageAttribute],
-    created: PreciseDateTime,
+    created: DateTime,
+    orderIndex: Int,
     var firstReceive: Received,
     var receiveCount: Int,
     isFifo: Boolean,
@@ -27,7 +27,10 @@ case class InternalMessage(
   override def compareTo(other: InternalMessage): Int = {
     if (isFifo) {
       // FIFO messages should be ordered on when they were written to the queue
-      -created.compareTo(other.created)
+      val comp = -created.compareTo(other.created)
+      if (comp == 0) {
+        -orderIndex.compareTo(other.orderIndex)
+      } else comp
     } else {
       // Plain messages are technically not ordered, but AWS seems to offer a loose, non-guaranteed "next delivery"
       // ordering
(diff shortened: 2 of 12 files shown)


adamw:
Thanks! Maybe, when sending a batch, we should artificially set the timestamps of subsequent messages increasing timestamps? E.g. adding the index in the batch to the nanoseconds or millis?

janjaali:
If I get you right you mean that we leave the `InternalMessage` precision as it was (i.e. `DateTime`) and decide at "receive-time" the creation time of the messages? I was thinking in the same direction at first but had two concerns with that solution: one was the artificially setting of an timestamp instead the precise one, actually we would get more vague about the creation time. The other concern was that the artificially setting of the message creation time has to be done at a time where we have the complete batch in our hand, and this is with the current implementation only possible at "receive-time". Both concerns together combined was getting me the feeling that we created a artificially, somehow vague, receive-timestamp instead of a more precise creation timestamp for the message. 

For a better discussion base, also maybe because my English is somewhat limited, I have created a branch to your suggestion, I hope this fits: https://github.com/softwaremill/elasticmq/compare/master...janjaali:receive-time-as-creation-time?expand=1.

I hope I get your suggestion right, otherwise please correct me and sorry for the long post :)

adamw:
No problem, thanks for the investigation! I think we might leave the improved message precision, however there's still a possibility that two messages get the same nano time. To be 100% sure that this will always work, I was thinking of adding the batch index to the nanoseconds - wouldn't that work?

janjaali:
Sure, we would have to add the maybe existing batch index to the `NewMessageData` and proxy it so to the `InternalMessage`'s:

```scala
case class NewMessageData(
    id: Option[MessageId],
    content: String,
    messageAttributes: Map[String, MessageAttribute],
    nextDelivery: NextDelivery,
    messageGroupId: Option[String],
    messageDeduplicationId: Option[String],
    batchIndex: Option[Int]
)
```

and 

```scala
case class PreciseDateTime(dateTime: DateTime, nanoSeconds: Long, batchIndex: Option[Int])

case class InternalMessage(
    ...
    created: PreciseDateTime,
    ...
)
```

The disadvantage would be to have a `batchIndex` or `orderingIndex` which is not traceable back to the original batch.

janjaali:
To hide the batchIndex in the `InternalMessage` we could simplify it by really "adding" it to the nanoseconds:

```scala
  def from(newMessageData: NewMessageData, queueData: QueueData): InternalMessage = {
    val now = System.currentTimeMillis()
    new InternalMessage(
      newMessageData.id.getOrElse(generateId()).id,
      mutable.Buffer.empty,
      newMessageData.nextDelivery.toMillis(now, queueData.delay.getMillis).millis,
      newMessageData.content,
      newMessageData.messageAttributes,
      {
        val preciseDateTime = PreciseDateTime()
        preciseDateTime.copy(nanoSeconds = preciseDateTime.nanoSeconds + newMessageData.batchIndex.get)
      },
      NeverReceived,
      0,
      queueData.isFifo,
      newMessageData.messageGroupId,
      newMessageData.messageDeduplicationId
    )
  }
```

adamw:
Or maybe we could add the `batchIndex` to `InternalMessage`? By default it would be `0`, and the sorting would be by the tuple `(timestamp, index)`?

janjaali:
how about calling it `orderIndex` instead since the internal messages know actually nothing about batches?

adamw:
👍

janjaali:
I have added an orderIndex to the internal messages, please have a re-review :)

adamw:
Released in 0.14.12, thanks!

PR #289, opened 2019-09-23
Make messageAttributeDigest optional

agwhitaker93:
There is currently an issue when using the PHP SDK, which expects a message to have message attributes if a value for MD5OfMessageAttributes is provided.


adamw:
Release in 0.14.14, thanks!

PR #294, opened 2019-10-04
Support string with multi-byte characters

yamachu:
ref: https://github.com/localstack/localstack/issues/1619

The above problem has revealed that localstack cannot handle multi-bytes characters.


adamw:
Released in 0.15.1, thanks!

PR #298, opened 2019-10-16
Set the default accountId to 12 digits of 0

yamachu:
Thanks to https://github.com/softwaremill/elasticmq/pull/290 , we can now set accountId and region by `.conf`.

But default accuntId is set 9 digits of 0, compatibility has been lost.
So I changed default accountId to 12 digits of 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ queues {
 // Region and accountId which will be included in resource ids
 aws {
     region = us-west-2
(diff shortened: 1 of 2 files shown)


adamw:
Released in 0.15.2, thanks!

yamachu:
Thank you for your quick reviewing and release!

PR #333, opened 2020-01-31
Handle tags when creating a queue

ImFlog:
Hello,

I found that when we create a queue via the rest api the tags are not processed.
I tried the fix using localstack and it seems to react as expected.
I couldn't find any test on this part of the code (but I am a newby in Scala / Akka), please help me if you want me to add some (even if it's a really small fix).

Thank you for the tool :)


adamw:
Released in 0.15.4. Thanks!

PR #340, opened 2020-03-04
Workaround. Added FifoQueue to ignore attribute list when setting queue attribute

tachocarpintero:


tachocarpintero:
The origin of the problem is explained here:
https://github.com/localstack/localstack/issues/2096

adamw:
Released in 0.15.5. Thanks!

PR #341, opened 2020-03-05
Add String MessageAttribute validation

yamachu:
According to MessageAttributeValue documents ( https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_MessageAttributeValue.html ) , String Value attribute must have the same restrictions on the content as the message body.

So added test and validator.

Message body restriction is written in link below.
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html


adamw:
Released in 0.15.5. Thanks!

PR #349, opened 2020-04-01
Improvements to GraalVM image generation

pbuda:
Replaces old, manual way of building both native image and a docker container containing it with sbt-native-packager way.


build.sbt
```diff
@@ -206,110 +204,51 @@ lazy val server: Project = (project in file("server"))
...
-val dockerGraalvmNativeImageName =
-  settingKey[String]("Name of the generated docker image, containing the native binary.")
-
 lazy val nativeServer: Project = (project in file("native-server"))
-  .enablePlugins(JavaServerAppPackaging, DockerPlugin)
+  .enablePlugins(GraalVMNativeImagePlugin, DockerPlugin)
   .settings(buildSettings)
   .settings(Seq(
     name := "elasticmq-native-server",
     libraryDependencies += "com.oracle.substratevm" % "svm" % "19.2.1" % Provided,
```
adamw, line 212: shouldn't this version match the `graalVMNativeImageGraalVersion`?

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.15.4"
+version in ThisBuild := "0.15.4.1"
```
adamw, line 1: this is changed automatically during a release

build.sbt
pbuda: It's interesting, because apparently there is a vm-20.0.0 tag on github, but the latest artifact in Maven Central is 19.2.1

version.sbt
pbuda: Fixed

adamw:
Released in 0.15.7, thanks!

PR #350, opened 2020-04-06
relax visibility of several queue actor methods to ease customisation

wigbam:
This PR is driven from a project where there was a need to provide a slightly customised Actor system, but which proved to be quite difficult due to strict visibility of methods in `Actor` and `QueueManagerActor` and thus resulted in quite a lot of copy-paste code.

Relaxing visibility of some methods in the above mentioned classes should allow for a better customisation support for those who wish to inherit from base implementation.


adamw:
Released in 0.15.6

PR #380, opened 2020-08-04
Add custom configuration to Dockerfile example

kayma-hl:
Cool project! I love it.

But as I had some problems to get my custom configuration file working within a docker container, the solution could be interesting for others too. So I added a few statements to the Dockerfile to make clear how it works.

Cheers
Mathias


adamw:
Thanks!

PR #397, opened 2020-10-09
Missing .fifo suffix for FIFO queues

### LINKED ISSUE #394: Fifo queue created with a config file does not end with `.fifo` suffix.
According to [SQS Docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html) "The name of a FIFO queue must end with the .fifo suffix.". But I cannot make it work with Elasticmq. 

When I define a custom config file `custom.conf` like this:

```
...
queues {
    testfoo {
        fifo = true
        defaultVisibilityTimeout = 5 minutes
        contentBasedDeduplication = true
    }
}
...
````

The service starts up. But the queue is named "testfoo", not "testfoo.fifo", even though it is fifo queue. Trying to connect to "testfoo" queue works, but "testfoo.fifo" ends with unknown queue error.

When I try to change the conf file and queue name to "testfoo.fifo" the service does not start. It looks like it treats this conf file as invalid.

The software I use depend on ".fifo" suffix to define if this is a SQS FIFO queue (well, you should be able to rely on that as the docs say "FIFO queue must end with the .fifo suffix."). But I cannot make it work.

I checked the test suite for elasticmq (I don't know much about Scala) and there seems to be a test https://github.com/softwaremill/elasticmq/blob/36cd45c2842588b9b92f63e910b9c81558069324/rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala#L427 for making sure fifo queue ends with ".fifo". And the next test makes queue pragramatically `new CreateQueueRequest("testQueue.fifo")`. And maybe it works when you create the queue with code but I have no way to check that. 

But for me it does not work with a conf file. Am I doing something wrong? Or is it a bug?

### PR DESCRIPTION
**Pull request by wojciechUrbanski**
Fixes #394. 

According to the documentation (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html), all FIFO queues should end up with .fifo suffix. 
Besides that, documentation mentions about 80 character queue name cap which also has been implemented as part of the #394 fix.

### CODE REVIEW

File: server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala [hunk 1]
```diff
@@ -9,6 +9,7 @@ import org.elasticmq.server.QueueSorter
 import org.elasticmq.util.Logging
 
 class ElasticMQServerConfig(config: Config) extends Logging {
```
adamw, line 11: This class only contains logic which parses configuration when the server is started & passes it to the actors, which manage the queues. The place where queues are managed & created is `QueueManagerActor` - and that's probably where name validation should go to.

This would probably become visible if you try to create a queue programatically - not through the config file, but through the SQS API. The config file is there to make it easier to setup tests, it's not the main interface for queue creation.

Try adding a test to `AmazonJavaSdkTestSuite` which tries to create a queue without the proper suffix, it will probably succeed (to create the queue, which means the test should fail :) ).

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
```diff
@@ -105,6 +107,20 @@ class ElasticMQServerConfig(config: Config) extends Logging {
     QueueSorter.sortCreateQueues(unsortedCreateQueues)
   }
 
+  private def resolveQueryName(nameFromConfig: String, isFifo: Boolean): String = {
+    if (isFifo) {
+      ensureLengthSmallerThan(nameFromConfig + ".fifo", MaximumQueueNameLength)
```
adamw, line 112: Should we add the `.fifo` suffix, or is it added automatically? How does the "real" SQS work - does it return an error, when trying to create a fifo queue with the wrong name, or does it add the suffix itself?

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
wojciechUrbanski: I have added it into the `ElasticMQServerConfig` because from the docs it seems, that FIFO queue without the ".fifo" suffix is something, that should not exist. Thus I have made an assumption, that it should be reported to user as fast as it is possible and abort the startup of the server (due to incorrect config file).

`AmazonJavaSdkTestSuite` already has a test `FIFO queues should return an error if the queue's name does not end in .fifo`, which is creating a FIFO queue without the ".fifo" suffix and the outcome is `AmazonServiceException`. So, Amazon Java SDK is not allowing to create a FIFO queue without the ".fifo" suffix.

adamw: But the configured queues are then created (synchronously) in `createQueues`, so any exceptions that happen during queue creation should be reported there anyway, no?

It seems that the Amazon Java SDK has a client-side check for this, but I would implement this in `QueueManagerActor` in case other clients (including non-official) don't have this check.

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
wojciechUrbanski: The "real" SQS will throw an exception if someone will try to create FIFO queue without the ".fifo" suffix.

Right now, if we will try to add ".fifo" suffix to the queue name in the config, server will throw an exception with message "fifo has type OBJECT rather than BOOLEAN" as the structure of the config has been changed. We are already using the "fifo" (as the property of queue) and expect it to be a boolean value.

The approach presented in this PR is based on the assumption, that users using the config file are relying on the already existing `fifo` property to indicate if the queue is FIFO, not on the `.fifo` suffix.

We could change this approach and enable the possibility to create queues with `.fifo` suffix in the config files. But I am not sure if it is correct approach to allow more than one way to achieve the same thing. It may be misleading for the end user if he can use either ".fifo" suffix or the "fifo" property. What do You think about it?

adamw: So using `name = "xyz.fifo"` will cause an exception being thrown by the config parser?

adamw: Ah because the queue name is not a value but a key in the config. Ok, let's add the `.fifo` to the queue name automatically, but also mention this in the readme on config (that this is automatically added).

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
wojciechUrbanski: Yeah, indeed it is better place for this validation. I have moved the queue name validation to `QueueManagerActor` and made small refactor of code related to SQS limits

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
```diff
@@ -5,13 +5,14 @@ import org.elasticmq.actor.queue.QueueActor
 import org.elasticmq.actor.reply._
 import org.elasticmq.msg.{CreateQueue, DeleteQueue, _}
 import org.elasticmq.util.{Logging, NowProvider}
-import org.elasticmq.{QueueAlreadyExists, QueueData}
+import org.elasticmq._
 
 import scala.reflect._
 
-class QueueManagerActor(nowProvider: NowProvider) extends ReplyingActor with Logging {
+class QueueManagerActor(nowProvider: NowProvider, sqsLimit: SQSLimits) extends ReplyingActor with Logging {
```
adamw, line 12: Maybe let's just call them: `Limits`. `SqsLimits` could then be the instantiation of `Limits` with values characteristic to `Sqs`.

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
```diff
@@ -5,13 +5,14 @@ import org.elasticmq.actor.queue.QueueActor
...
   type M[X] = QueueManagerMsg[X]
   val ev: ClassTag[QueueManagerMsg[Unit]] = classTag[M[Unit]]
+  val MaximumQueueNameLength = 80
```
adamw, line 15: this is now unused?

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
```diff
@@ -23,9 +24,19 @@ class QueueManagerActor(nowProvider: NowProvider) extends ReplyingActor with Log
           Left(new QueueAlreadyExists(queueData.name))
         } else {
           logger.info(s"Creating queue $queueData")
-          val actor = createQueueActor(nowProvider, queueData)
-          queues(queueData.name) = actor
-          Right(actor)
+          for {
```
adamw, line 27: [minor] I think this would be more readable by explicitly handling left/rigth:

```
Limits.verifyQueueName(...) match {
  case Left(msg) => Left(QCE(...))
  case Right(_) => val actor = ...
}
```

core/src/main/scala/org/elasticmq/SQSLimits.scala
```diff
@@ -0,0 +1,127 @@
+package org.elasticmq
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+sealed trait SQSLimits {
```
adamw, line 6: probably doesn't have to be sealed, very unlikely but we could without problems accept other implementations, no?

core/src/main/scala/org/elasticmq/SQSLimits.scala
```diff
@@ -0,0 +1,127 @@
...
+      "AWS.SimpleQueueService.TooManyEntriesInBatchRequest"
+    )
+
+  def verifyNumberOfMessagesFromParameters(
+      numberOfMessagesFromParameters: Int,
+      sqsLimit: SQSLimits
+  ): Either[String, Unit] =
+    validateWhenLimitAvailable(sqsLimit.numberOfMessagesLimit)(
+      limit => limit.isBetween(numberOfMessagesFromParameters),
+      "ReadCountOutOfRange"
```
adamw, line 68: these error messages don't have the `AWS.SimpleQueueService.` prefix like the above?

adamw:
👍 Looks good, just a couple usually minor comments

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
wojciechUrbanski: Changed them to `Limits` but I have not changed the structure of it yet. I propose to stay with how it is done right now and if someone will have another use case, then it could be remodelled according to the needs.

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
wojciechUrbanski: Yup, I have missed it, removed

core/src/main/scala/org/elasticmq/SQSLimits.scala
wojciechUrbanski: There is possibility that there will be another implementations so I have "unsealed" it

core/src/main/scala/org/elasticmq/SQSLimits.scala
wojciechUrbanski: Nope, but I am not sure why only some of them have this prefix and the rest does not. I didn't want to introduce any additional changes in this PR so I have left it as it was.

adamw:
No more comments now except a test failing on 2.13 :)

santialbo:
Please ignore this comment, I just want google to index it for people who had the same issue as me.

If you suddenly started getting `AWS.SimpleQueueService.NonExistentQueue` out of the blue and you are using fifo queues, make sure that the elasticmq endpoint/queueUrl you are passing to the aws client contains the `.fifo` suffix.

PR #406, opened 2020-10-22
Validate empty string attributes

### LINKED ISSUE #373: The behavior on sending message-attributes with empty value is different from that of SQS
When I send a message to sqs with empty value, sqs returns error (InvalidParameterValue).
```
cat attributes.json
{
  "piyo": {
    "DataType": "String",
    "StringValue": ""
  }
}
aws sqs send-message \
    --queue-url https://sqs.ap-northeast-1.amazonaws.com/XXXXXXXXXX/elastic-mq-debug \
    --message-body bar \
    --message-attributes file://attributes.json
An error occurred (InvalidParameterValue) when calling the SendMessage operation: Message (user) attribute 'piyo' must contain a non-empty value of type 'String'.
```

But elastmq does not return error.

```
 java -jar elasticmq-server-0.15.7.jar
...

aws --endpoint-url http://localhost:9324 sqs create-queue --queue-name foo
{
    "QueueUrl": "http://localhost:9324/queue/foo"
}
cat attributes.json
{
  "piyo": {
    "DataType": "String",
    "StringValue": ""
  }
}
aws --endpoint-url http://localhost:9324 \
    sqs send-message \
    --queue-url http://localhost:9324/queue/foo \
    --message-body bar \
    --message-attributes file://attributes.json
{
    "MD5OfMessageBody": "37b51d194a7513e45b56f6524f2d51f2",
    "MD5OfMessageAttributes": "4166cc16b73099be3abe8e2b0f544c99",
    "MessageId": "62aa4db3-9666-4421-a1d8-b988f76b7ae1"
}
echo $?
0
```

wojciechUrbanski:
Fixes #373 

According to doc, message attribute value/data type can't be an empty String. Same applies to message body

### DIFF
--- a/core/src/main/scala/org/elasticmq/Limits.scala
+++ b/core/src/main/scala/org/elasticmq/Limits.scala
@@ -72,8 +72,8 @@ object Limits {
     )
 
   def verifyMessageStringAttribute(
-      attributeValue: String,
       attributeName: String,
+      attributeValue: String,
       limits: Limits
   ): Either[String, Unit] =
     for {
--- a/core/src/main/scala/org/elasticmq/Limits.scala
+++ b/core/src/main/scala/org/elasticmq/Limits.scala
@@ -7,6 +7,7 @@ trait Limits {
   def queueNameLengthLimit: Limit[Int]
   def batchSizeLimit: Limit[Int]
   def numberOfMessagesLimit: Limit[RangeLimit[Int]]
+  def nonEmptyAttributesLimit: Limit[Boolean]
   def bodyValidCharactersLimit: Limit[List[RangeLimit[Int]]]
   def numberAttributeValueLimit: Limit[RangeLimit[BigDecimal]]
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/elasticmq/Limits.scala
```diff
@@ -68,25 +71,57 @@ object Limits {
       "ReadCountOutOfRange"
     )
 
-  def verifyMessageStringAttribute(stringAttribute: String, limits: Limits): Either[String, Unit] =
+  def verifyMessageStringAttribute(
+      attributeValue: String,
+      attributeName: String,
```
adamw, line 76: shouldn't it be name-value, as that's the usual order? :)

adamw (approved):

core/src/main/scala/org/elasticmq/Limits.scala
wojciechUrbanski: Yeah, I have changed it to name-value order :)

PR #409, opened 2020-10-28
Loading custom config files in elasticmq native

### LINKED ISSUE #368: elasticmq native does not load /opt/elasticmq.conf file
It seems that `elasticmq-native` ignores the ` /opt/elasticmq.conf` file, making it impossible to pre-configure queues before start-up.

wojciechUrbanski:
Fixes #368 

With those changes it is possible to start the docker with mounted custom config via command `docker run --rm -it --publish 9324:9324 -v /path/to/elasticmq.conf:/opt/elasticmq.conf softwaremill/elasticmq-native`. Before, the path was `/opt/docker/elasticmq.conf` which was not compliant with standard elasticmq image.

I have also added few classes to `reflect-config.json` as the application was failing with `NoClassDefFoundError`


mchmielarz:
👍

mchmielarz (commented):
👍

yanns:
Thx a lot for the fix!
I'm now using the native version. ❤️

PR #412, opened 2020-11-03
Deduplication logic

### LINKED ISSUE #354: fifo queues do not deduplicate within 5 minutes if message is already deleted
Deduplication only works if the duplicate message is still in the queue.

There is a TODO on this in the code, but I don't see an issue on it.
https://github.com/softwaremill/elasticmq/blob/master/core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
```
// Ensure a message with the same deduplication id is not on the queue already. If the message is already on the
// queue, return that -- don't add it twice
// TODO: A message dedup id should be checked up to 5 mins after it has been received. If it has been deleted
// during that period, it should _still_ be used when deduplicating new messages. If there's a match with a
// deleted message (that was sent less than 5 minutes ago, the new message should not be added).
```

wojciechUrbanski:
Fixes #354 

Implemented solution:

On the Queue level I have added new structure `FifoDeduplicationIdsHistory`. Main responsibility of it, is to store deduplication IDs that were received on `SendMessage`. History has two fast access storages:
* First one is used for checking if given Deduplication ID has been already used
* Second one purpose is to store deduplication IDs together with `created` date in oldest->newest sort order. It is used for cleaning up obsolete IDs.
Each Queue will start their own timer which will each second check if any Deduplication ID is obsolete (created date exceeded 5 minutes mark). If so, it is removed from the History.

Minor reworks/adjustments:
* Dividing `QueueActorMessageOps` to smaller traits, each responsible for one action
* Added DeduplicationId domain object
* Extracting tests related to deduplication from `AmazonJavaSdkTestSuite` and moving them to `FifoDeduplicationTests`
* Added logic responsible for overwriting deduplication ID of message once it is moved to DLQ. Based on the real SQS, deduplication ID should be changed to message ID once it is moved to DLQ.


core/src/main/scala/org/elasticmq/FifoDeduplicationIdsHistory.scala
```diff
@@ -0,0 +1,74 @@
...
+      case head :: tail if cond(head) => go(mapppedElements :+ map(head), tail)
+      case _ :: _                     => (mapppedElements, restOfTheColl)
+      case Nil                        => (mapppedElements, List.empty)
+    }
+    go(List.empty, coll)
+  }
+}
+
+object FifoDeduplicationIdsHistory {
+  val deduplicationInterval = 5
```
adamw, line 64: Would be good to indicate the unit (minutes). So either name it `DeduplicationIntervalMinutes`, or maybe store it as a java/scala `Duration`?

core/src/main/scala/org/elasticmq/FifoDeduplicationIdsHistory.scala
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  def wasRegistered(maybeDeduplicationId: Option[DeduplicationId]): Option[InternalMessage] =
+    maybeDeduplicationId.flatMap(deduplicationId => messagesByDeduplicationId.get(deduplicationId))
+
+  def cleanOutdatedMessages(nowProvider: NowProvider): FifoDeduplicationIdsHistory = {
+    val (idsToRemove, notTerminatedMessages) =
+      partitionMapUntil(deduplicationIdsByCreationDate)(
+        map = _.id,
+        cond = _.creationDate.plusMinutes(deduplicationInterval).isBefore(nowProvider.now)
```
adamw, line 40: [minor] we could capture the `now` once, to avoid getting the system time on each loop

core/src/main/scala/org/elasticmq/actor/queue/operations/CommonOperations.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.elasticmq._
+import org.elasticmq.actor.queue.InternalMessage
+import org.elasticmq.util.NowProvider
+
+object CommonOperations {
+
+  def wasRegistered(
+      newMessage: NewMessageData,
+      deduplicationIdsHistory: FifoDeduplicationIdsHistory
+  ): Option[InternalMessage] = {
```
adamw, line 12: is this ever used (the `InternalMessage`)? maybe it would be sufficient to store a `Set[DeduplicationId]`, instead of a map to messages?

core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
```diff
@@ -1,211 +1,40 @@
...
+    with DeleteMessageOps
+    with ReceiveMessageOps
+    with MoveMessageOps
+    with Timers {
   this: QueueActorStorage =>
 
   def nowProvider: NowProvider
   def context: akka.actor.ActorContext
 
+  timers.startTimerWithFixedDelay(s"Timer: ${queueData.name}", DeduplicationIdsCleanup, 1.second)
```
adamw, line 24: shouldn't this be done in actor's `postStart` or `preStart` or some callback like that?

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/FifoDeduplicationTests.scala
```diff
@@ -0,0 +1,232 @@
...
+import org.elasticmq.rest.sqs.model.RedrivePolicy
+import org.elasticmq.rest.sqs.model.RedrivePolicyJson.format
+import org.scalatest.OptionValues
+import org.scalatest.matchers.should.Matchers
+import spray.json.enrichAny
+
+import scala.collection.JavaConverters.mapAsJavaMapConverter
+import scala.jdk.CollectionConverters.CollectionHasAsScala
+
+class FifoDeduplicationTests extends SqsClientServerCommunication with Matchers with OptionValues {
```
adamw, line 18: do we test that the dedup ids are removed after 5 minutes? (by shifting the `NowProvider`, of course)

adamw (approved):

core/src/main/scala/org/elasticmq/actor/queue/operations/CommonOperations.scala
wojciechUrbanski: It is used in `SendMessage` action. If we find that the message is duplicated, we want to return the old one (with old message ID etc.)

core/src/main/scala/org/elasticmq/actor/queue/QueueActorMessageOps.scala
wojciechUrbanski: It is already done in the `Timers` trait

core/src/main/scala/org/elasticmq/actor/queue/operations/CommonOperations.scala
adamw: Ah ok, missed that :)

core/src/main/scala/org/elasticmq/FifoDeduplicationIdsHistory.scala
wojciechUrbanski: I have changed it to `DeduplicationIntervalMinutes` as the place where it is used needs just an `Int`

core/src/main/scala/org/elasticmq/FifoDeduplicationIdsHistory.scala
wojciechUrbanski: Yeah, I even thought about it and put it into "backlog" of things to do, but in the end forgot about it :P Changed to invoke `now` only once

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/FifoDeduplicationTests.scala
wojciechUrbanski: I have added a test for this scenario :)

PR #424, opened 2020-11-16
REST server for queues statistics 

dbroda:
It is an expansion of statistics gathered internally. On the one hand, they are served as mbeans (on branch ui-mbeans). On the second hand, these commits expose REST endpoints with queues statistics and their attributes.


rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsDirectives.scala
```diff
@@ -0,0 +1,97 @@
...
+  implicit val queueFormat = jsonFormat2(QueueResponse)
+}
+
+trait StatisticsDirectives extends JsonSupport {
+  this: ElasticMQDirectives with QueueAttributesOps with NowProviderHolder =>
+
+  lazy val ec: ExecutionContext = ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(2))
+  implicit val duration = timeout.duration
+
+  val np: NowProvider = nowProvider
```
wojciechUrbanski, line 48: Preference opinion:
For me, it is clearer to pass dependencies explicitly instead of depending on the inheritance.
For example, in `StatisticsRestServerBuilder` fact that Statistics endpoints require dependency on time abstraction is hidden. On the other hand, if `statistics` method would require explicitly `NowProvider` then it is visible at first glance that there is such dependency in this functionality.

core/src/main/scala/org/elasticmq/util/NowProviderHolder.scala
```diff
@@ -0,0 +1,6 @@
+package org.elasticmq.util
+
+trait NowProviderHolder {
+
+  val nowProvider: NowProvider = new NowProvider
```
wojciechUrbanski, line 5: Maybe it would be better to create one abstraction without any behaviour:
```scala
trait NowProviderHolder {
    def nowProvider: NowProvider
}
```

and then instantiate two independent concrete implementations:
```scala
trait MutableNowProviderHolder extends NowProviderHolder { ... }
trait RealNowProviderHolder extends NowProviderHolder { ... }
```

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsDirectives.scala
dbroda: Yeah, this will be much simpler. Thanks

core/src/main/scala/org/elasticmq/util/NowProviderHolder.scala
dbroda: I've removed holders, NowProviders are passed explicitly

rest/rest-sqs/src/test/scala/org/elasticmq/rest/stats/StatisticsDirectivesTest.scala
```diff
@@ -0,0 +1,109 @@
...
+
+    Get("/statistics/queues") ~> route ~> check {
+      responseAs[List[QueuesResponse]] should contain theSameElementsAs List(
+        QueuesResponse("firstQueue", QueueStatisticsResponse(0, 0, 0)),
+        QueuesResponse("secondQueue", QueueStatisticsResponse(0, 0, 0))
+      )
+    }
+  }
+
+  "statisticsRequestForCertainQueue" should "return all queues statistics" in {
```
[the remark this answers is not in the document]
dbroda: Fixed, thanks

core/src/main/scala/org/elasticmq/util/MutableNowProviderHolder.scala
```diff
@@ -0,0 +1,10 @@
+package org.elasticmq.util
+
+import org.joda.time.DateTime
+
+trait MutableNowProviderHolder extends NowProviderHolder {
```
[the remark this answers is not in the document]
dbroda: It was removed after review.

core/src/main/scala/org/elasticmq/util/MutableNowProvider.scala
```diff
@@ -1,12 +1,15 @@
...
 
-import org.joda.time.DateTime
-import org.elasticmq.util.NowProvider
 import java.util.concurrent.atomic.AtomicLong
 
-class MutableNowProvider extends NowProvider {
-  var mutableNowMillis = new AtomicLong(100L)
+import org.joda.time.DateTime
+
+class MutableNowProvider(initalValue: Long = 100L) extends NowProvider {
```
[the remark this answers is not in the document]
dbroda: I've moved it back into tests package. Also, sqsRest-test has a dependency on "core % test" module

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
```diff
@@ -19,9 +20,11 @@ class ElasticMQServer(config: ElasticMQServerConfig) extends Logging {
   def start() = {
     val queueManagerActor = createBase()
     val restServerOpt = optionallyStartRestSqs(queueManagerActor)
+    val restStatisticsServerOpt = optionallyStartRestStatistics(queueManagerActor)
 
     val shutdown = () => {
       restServerOpt.map(_.stopAndGetFuture())
+      restStatisticsServerOpt.map(_.stopAndGetFuture())
```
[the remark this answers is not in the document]
adamw: Yes, this probably should be sth like `unbindRestServer.flatMap(_ => unbinedStatsServer.flatMap(_ => actorSystem.terminate()))`, and then we can await on this future. Probably can be written down in a nicer way using for-comprehensions :)

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/QueueAttributesOps.scala
```diff
@@ -0,0 +1,197 @@
...
+import org.elasticmq.util.Logging
+import org.elasticmq.{DeadLettersQueueData, MillisVisibilityTimeout, QueueData}
+import org.joda.time.Duration
+import spray.json.JsonParser.ParsingException
+import spray.json._
+
+import scala.async.Async.{async, await}
+import scala.concurrent.{ExecutionContext, Future}
+
+trait QueueAttributesOps extends AttributesModule {
```
adamw, line 20: is this only a refactor from `QueueAttributesDirectives` or are there some changes?

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsDirectives.scala
```diff
@@ -0,0 +1,95 @@
...
+trait JsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
+  implicit val queueStatisticsFormat = jsonFormat3(QueueStatisticsResponse)
+  implicit val queuesFormat = jsonFormat2(QueuesResponse)
+  implicit val queueFormat = jsonFormat2(QueueResponse)
+}
+
+trait StatisticsDirectives extends JsonSupport {
+  this: ElasticMQDirectives with QueueAttributesOps  =>
+
+  lazy val ec: ExecutionContext = ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(2))
```
adamw, line 45: we can probably reuse the ec from akka. I think there's a directive - `extractExecutionContext`, or sth like that

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsDirectives.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  def gatherAllQueuesWithStats(implicit np: NowProvider): Future[Iterable[QueuesResponse]] = {
+
+    QueueMetricsOps.getQueuesStatistics(queueManagerActor, np)
+      .map { x => x.map { case (name, stats) => mapToRestQueuesResponse(name, stats) } }
+  }
+
+  def gatherSpecificQueueWithAttributesRoute(queueName: String): Route = {
+    val map = Map("QueueName" -> queueName)
+    queueActorAndDataFromRequest(map) { (queueActor, queueData) =>
```
adamw, line 71: maybe this can be done simpler - looking  up the actor by name?

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
```diff
@@ -65,6 +68,28 @@ class ElasticMQServer(config: ElasticMQServerConfig) extends Logging {
     }
   }
 
+  private def optionallyStartRestStatistics(queueManagerActor: ActorRef): Option[StatisticsRestServer] = {
+    if (config.restStatisticsConfiguration.enabled) {
+
+      val server = TheStatisticsRestServerBuilder(
+        Some(actorSystem),
+        Some(queueManagerActor),
```
adamw, line 76: since these are always `Some` we can probably simplify by using non-optional parameters here (I know it's copied from the sqs server, but the optionals over there are a historical artifact, probably can also be removed, but that's another story :) )

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/package.scala
```diff
@@ -0,0 +1,6 @@
+package org.elasticmq.rest
+
+package object stats {
+  type Attributes = Map[String, String]
```
adamw, line 4: is this used?

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/QueueAttributesOps.scala
dbroda: refactor only - I've extracted logic here to reuse it in statistics module

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/package.scala
dbroda: Yes, in case-class returned in rest endpoint.

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
dbroda: Sure, I'll take care of that ;)

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsDirectives.scala
dbroda: I'll try it! Thanks!

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
dbroda: I've refactored TheStatisticsRestServerBuilder to not use optional actor system. I've looked also on SQSRestServerBuilder and have some doubts about changing that one. I could force to use non-optional value, but It could break some backward compatibility if anyone used it already. We are using that's None possibility on some tests.
@adamw do you think that changing only Statistics builder will be enough?

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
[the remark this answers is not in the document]
dbroda: I consider implementing it this way:

```
   val shutdown = () => {
      implicit val ec: ExecutionContextExecutor = actorSystem.dispatcher

      val futureTerminationRestSQS = restServerOpt.map(_.stopAndGetFuture()).getOrElse(Future.unit)
      val futureTerminationRestStats = restStatisticsServerOpt.map(_.stopAndGetFuture()).getOrElse(Future.unit)
      val eventualTerminated = for {
        _ <- futureTerminationRestSQS
        _ <- futureTerminationRestStats
        ac <- actorSystem.terminate()
      } yield ac
      Await.result(eventualTerminated, Inf)
    }
```

What do you think guys?

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
adamw: Yes, let's leave `SQSRestServerBuilder` for another occasion. That's not the scope of this task, better to keep PRs small :)

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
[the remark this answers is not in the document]
adamw: looks good! :)

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/package.scala
adamw: ok :)

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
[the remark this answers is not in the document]
dbroda: grazie! :)

server/src/main/scala/org/elasticmq/server/ElasticMQServer.scala
dbroda: 👍

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsDirectives.scala
dbroda: I've rechecked its usage. Function queueActorAndDataFromRequest gives also "queue data" that is used in the logic of gathering attributes. When I'll try to lookup for actors by name I will need to implement this exactly like the "queueActor..." method does. So I coming to the conclusion that maybe I should leave it that way like it is... ;) What do you think?

dbroda: OK got your point. I've rewritten it without using a map. Thanks!

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsRestServerBuilder.scala
```diff
@@ -0,0 +1,159 @@
...
+import org.elasticmq.rest.sqs.directives.ElasticMQDirectives
+import org.elasticmq.util.{Logging, NowProvider}
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+import scala.util.control.NonFatal
+
+case class TheStatisticsRestServerBuilder(
+                                           providedActorSystem: ActorSystem,
+                                           providedQueueManagerActor: Option[ActorRef],
```
adamw, line 22: maybe it's possible to remove this option as well?

adamw (approved):

rest/rest-sqs/src/main/scala/org/elasticmq/rest/stats/StatisticsRestServerBuilder.scala
dbroda: 👍

wojciechUrbanski (approved):

PR #441, opened 2021-01-14
GetQueueURL should return url containing aws id, not queue keyword - fixes #421

### LINKED ISSUE #421: GetQueueUrl should return aws account id if set
The current GetQueueUrl implementation is not fully compatible with the SQS interface.

The response returns "queue" where the awsAccountId should be.

 ie:
`{protocol}://{host}/queue/{queue_name}`

instead of:
`{protocol}://{host}/{account_id}/{queue_name}`

This is always set to `QueueUrlContext` here: https://github.com/softwaremill/elasticmq/blob/28ef9c95126e3fdca950a6c98df8ed534b0be9e3/rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala#L405

Get Queue Url docs: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueUrl.html

Happy to work on a PR if there's no reason why this couldn't be changed.

michalo2882:


adamw:
Is writing a test possible? :)

michalo2882:
there is one already, but the check wasn't specific enough. is it ok now?

adamw:
Yes, looks good. One more question though - does this change the default behavior in some way? That is, will the queue urls change if somebody used the default configuration?

michalo2882:
Yes, they will change, because the default configuration contains some aws id. However, both `queue/<queue_name>` and `<aws_id>/<queue_name>` urls work for the existing endpoints, so technically the change doesn't break compatibility.

adamw:
Ok, we'll release it as version 1.1 then :)

PR #443, opened 2021-01-15
delete msg - check if receipt handle is valid - fixes #324

### LINKED ISSUE #324: delete-message with invalid receipt handle doesn't throw any error
Assume we have two queues, one is in the AWS and the other resides in the ElasticMQ locally. Both are empty.

When I try to delete the message from the queue in the AWS using the invalid receipt handle, I get the following:
```
aws sqs delete-message --queue-url https://queue.amazonaws.com/real-queue-url --receipt-handle fake

An error occurred (ReceiptHandleIsInvalid) when calling the DeleteMessage operation: The input receipt handle "fake" is not a valid receipt handle.
```

And when I try to do the same operation on the local queue in the elasticMQ it performs silently:
```
aws --endpoint-url http://localhost:9324 sqs delete-message --queue-url http://localhost:9324/queue/local_queue --receipt-handle fake
```

I start elasticMQ from Docker container by running `docker run -p 9324:9324 softwaremill/elasticmq` 
The version of the image is `0.15.3`

michalo2882:


core/src/main/scala/org/elasticmq/ElasticMQError.scala
```diff
@@ -20,3 +20,8 @@ class MessageDoesNotExist(val queueName: String, messageId: MessageId) extends E
   val code = "MessageDoesNotExist"
   val message = s"Message does not exist: $messageId in queue: $queueName"
 }
+
+class ReceiptHandleIsInvalid(val queueName: String, receiptHandle: String) extends ElasticMQError {
+  val code = "ReceiptHandleIsInvalid"
+  val message = s"""The input receipt handle "$receiptHandle" is not a valid."""
```
adamw, line 26:
```suggestion
  val message = s"""The receipt handle "$receiptHandle" is not valid."""
```

core/src/main/scala/org/elasticmq/ElasticMQError.scala
```diff
@@ -20,3 +20,8 @@ class MessageDoesNotExist(val queueName: String, messageId: MessageId) extends E
   val code = "MessageDoesNotExist"
   val message = s"Message does not exist: $messageId in queue: $queueName"
 }
+
+class ReceiptHandleIsInvalid(val queueName: String, receiptHandle: String) extends ElasticMQError {
```
adamw, line 24: maybe simply: `InvalidReceiptHandle`?

michalo2882: Fixed, although I left the string used in the api error response to be `ReceiptHandleIsInvalid`, so it conforms to https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_DeleteMessage.html

PR #445, opened 2021-01-18
Embedded server custom config

michalo2882:


adamw:
I'd keep the readme, but drop the config changes in the server - there isn't really any utility in providing custom akka configuration (other than confusion ;) )

adamw:
Thanks!

PR #487, opened 2021-05-20
Added dockerBaseImage supporting arm64 architecture and instruction in readme

rojekp:


README.md
```diff
@@ -396,6 +396,15 @@ Generating GraalVM config files is a manual process currently. You need to run t
...
 These files should be placed in `native-server/src/main/resources/META-INF/native-image` and are automatically used by the native-image process.
 
+## Building multi-architecture image
+
+Currently docker image supports one platform: `amd64`. To publish it for two different platforms: `amd64` and `arm64`, follow the instruction (macOS).
+
+* Generate the Dockerfile using SBT - executing `sbt docker:stage` will generate it in `target/docker/stage`
+* Check docker buildx `docker buildx version`
+* Create a new builder instance and enable it `docker buildx create --use --name cross-platform-builder` - by default, buildx is using `docker` driver but we need `docker-container` driver instead.
+* Generate image and push to Docker Hub: `docker buildx build --platform=linux/arm64,linux/amd64 --push -t softwaremill/elasticmq:<tag> .`
```
adamw, line 406: can't we perform these steps when releasing our docker images?

rojekp: Pushed new commit with automatic support and updated readme.

build.sbt
```diff
@@ -212,7 +214,13 @@ lazy val server: Project = (project in file("server"))
       credentials += Credentials(Path.userHome / ".s3_elasticmq_credentials"),
       // docker
       dockerExposedPorts := Seq(9324,9325),
-      dockerBaseImage := "openjdk:8u212-b04-jdk-stretch",
+      dockerBaseImage := "openjdk:11-jdk-stretch",
+      dockerBuildOptions := dockerBuildOptions.value :+ "--platform=linux/arm64,linux/amd64" :+ "--push",
+      dockerBuildCommand := {
+        val old = dockerBuildCommand.value
+        val withBuildx = old.take(1) ++ Seq("buildx") ++ old.drop(1)
```
adamw, line 221: some comments would be good explaining why we replace that element - what is the structure of the command.

also, this assumes that the returned list has always 3 elements. Wouldn't it be safer to replace a specific element, and fail if there are e.g. 4 elements?

README.md
```diff
@@ -396,6 +396,24 @@ Generating GraalVM config files is a manual process currently. You need to run t
 
 These files should be placed in `native-server/src/main/resources/META-INF/native-image` and are automatically used by the native-image process.
 
+## Building multi-architecture image
+
+Currently, docker image supports one platform: `amd64`. To publish it for two different platforms: `amd64` and `arm64`, module `server` in `build.sbt` is modified as follows:
```
adamw, line 401: after merging, "currently" will become "in the past". The readme should explain the current setup, not the history of changes - that's what e.g. commit messages, PR messages, release notes are for.

So here we need an explanation on how the build process works, how to set it up locally (do I need to install the buildx plugin somehow?), and how to build locally for a single architecture

rojekp: Ok, I modified readme

build.sbt
rojekp: Added comment, should be clear enough. The returned list will not have always 3 elements. It may be more or less and it's safe. We only need command `buildx` after first `[dockerExecCommand]` which is docker by default and there is no need to modify that exec command.

adamw: ah ok, you want to add, not replace an element. Clear now with the comment :)

build.sbt
```diff
@@ -212,7 +214,15 @@ lazy val server: Project = (project in file("server"))
       credentials += Credentials(Path.userHome / ".s3_elasticmq_credentials"),
       // docker
       dockerExposedPorts := Seq(9324,9325),
-      dockerBaseImage := "openjdk:8u212-b04-jdk-stretch",
+      dockerBaseImage := "openjdk:11-jdk-stretch",
```
adamw, line 217: shouldn't we do the same for `nativeServer`?

rojekp: According to comments in the issue - `nativeServer` works on ARM-based machines.

PR #488, opened 2021-05-28
Added support for building multi-arch Docker images for elasticmq-native

rojekp:


build.sbt
```diff
@@ -283,6 +287,14 @@ lazy val nativeServer: Project = (project in file("native-server"))
     dockerEntrypoint := Seq("/sbin/tini", "--", "/opt/docker/bin/elasticmq-native-server", "-Dconfig.file=/opt/elasticmq.conf"),
     dockerUpdateLatest := true,
     dockerExposedPorts := Seq(9324,9325),
+    dockerBuildOptions := dockerBuildOptions.value :+ "--platform=linux/arm64,linux/amd64" :+ "--push",
+    dockerBuildCommand := {
```
adamw, line 291: maybe this can be part of `dockerBuildxSettings` and hence not repeated?

build.sbt
```diff
@@ -349,7 +364,7 @@ lazy val uiSettings = Seq(
 lazy val dockerBuildxSettings = Seq(
   createBuildx := {
     streams.value.log("Creating docker buildx instance")
-    haltOnCmdResultError(Process("docker buildx create --use --name multi-arch-builder", baseDirectory.value).!)
+    Process("docker buildx create --use --name multi-arch-builder", baseDirectory.value).!
```
adamw, line 367: we don't want to halt? :)

build.sbt
rojekp: Moved that out like suggested.

build.sbt
rojekp: Well, that command will return either success with new Buildx instance or just info about existing instance.

adamw: So if the instance exist, the command ends in error?

PR #491, opened 2021-05-31
Fixing releasing process - refactoring docker multi-arch building process

rojekp:


build.sbt
```diff
@@ -224,8 +222,18 @@ lazy val server: Project = (project in file("server"))
       ),
       publishLocal in Docker := (publishLocal in Docker)
         .dependsOn(yarnTask.toTask(" build"))
-        .dependsOn(createBuildx)
         .value,
+      publish in Docker := {
+        val _ = publishLocal.value
```
adamw, line 227: try: `(publishLocal in Docker).value`

rojekp: It works, I had to fix appropriate directory also.

build.sbt
```diff
@@ -305,8 +312,17 @@ lazy val nativeServer: Project = (project in file("native-server"))
     dockerUsername := Some("softwaremill"),
     packageBin in GraalVMNativeImage := (packageBin in GraalVMNativeImage)
       .dependsOn(yarnTask.toTask(" build"))
-      .dependsOn(createBuildx)
       .value,
+    publish in Docker := {
```
adamw, line 316: aren't the above the same? If so, we should extract them to a settings list like `dockerBuildxSettings`

rojekp: Ok, extracted.

PR #493, opened 2021-06-01
Changed exception when sending msg to queue without proper url

rojekp:


rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/package.scala
```diff
@@ -22,5 +22,14 @@ package object sqs {
         case None    => reject(MissingQueryParamRejection(n))
       }
     }
+
+    def requiredUrlParam(n: String): Directive1[String] = {
+      p.get(n) match {
+        case Some(v) => provide(v)
+        case None    => reject(WrongURLFormatRejection(n))
```
adamw, line 29: hm well if the parameter is absent, then the format isn't wrong, it's missing, no? :)

rojekp: Yes, changed back to `MissingQueryParamRejection` :)

rojekp: Yes, resulting exception is still the same.

adamw: I mean, the method `requiredUrlParam` gives a `WrongURLFormatRejection` which is not very logical

rojekp: Modified solution, added new directive.

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/directives/QueueDirectives.scala
```diff
@@ -76,4 +77,18 @@ trait QueueDirectives {
       body(queueData)
     }
   }
+
+  private def checkOnlyQueueNameInUri(): Directive1[String] = {
+    extractUri.flatMap {
+      uri => {
+        val onlyQueueNameInUriPath = Option(uri.path.toString().tail)
```
adamw, line 84: doesn't this always return `Some(the uri without the first character)`?

rojekp: Thanks for advice, now I'm checking if there is only one segment in uri.

Issue #511, opened 2021-07-22
NoSuchElementException when tag on SQS is set to nil


frncmx-form3:
We use ElasticMQ via Localstack and that is provisioned with Terraform. Basically, the same way as we provision our real AWS infra. Recently, I came across a weird edge case when one of our TF module set an SQS tag to empty string, i.e., 

```terraform
resource "aws_sqs_queue" "queue" {
  name = "foobar"

  tags = {
    NoFoo = ""
  }
}
```
As that travels through the AWS TF provider and Localstack, somehow it reaches ElasticMQ as `nil` not as an empty string and that causes an exception. Plus the resource creation hangs. I'm not sure how and when the empty string turns into `nil`. However, maybe ElasticMQ should handle this edge case gracefully. So I decided to post this issue. 

## How to reproduce

I run ElasticMQ as `docker run --rm -p 9324:9324 softwaremill/elasticmq:1.1.1`

The way I reproduced the issue is with the following Go snippet:

```go
package main

import (
	"fmt"
	"net/http/httputil"
	"time"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/request"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/sqs"
)

const elasticMQ = "http://localhost:9324"

func main() {
	client := newSQSClient()

	req, resp := client.CreateQueueRequest(&sqs.CreateQueueInput{
		Attributes: nil,
		QueueName:  aws.String(fmt.Sprintf("foobar-%d", time.Now().Unix())),
		Tags: map[string]*string{
			//"some": aws.String("foo"), // OK
			//"empty": aws.String(""), // OK
			"NoFoo": nil, // Exception
		},
	})

	panicOnErr(req.Build())
	dump(req)

	panicOnErr(req.Send())
	fmt.Println("New queue created with URL", *resp.QueueUrl)
}

func dump(req *request.Request) {
	result, err := httputil.DumpRequestOut(req.HTTPRequest, true)
	panicOnErr(err)

	fmt.Println("#####################################")
	fmt.Println(string(result))
	fmt.Println("#####################################")
}

func newSQSClient() *sqs.SQS {

	sess := session.Must(session.NewSessionWithOptions(session.Options{
		Config: aws.Config{
			Endpoint: aws.String(elasticMQ),
			Region:   aws.String("elasticmq"),
		},
	}))
	return sqs.New(sess)
}

[...]

PR #513, opened 2021-07-27
Changed way of reading tags without values

bartekzylinski:
closing #511


rest/rest-sqs/src/test/scala/org/elasticmq/rest/sqs/TagsTest.scala
```diff
@@ -0,0 +1,13 @@
+package org.elasticmq.rest.sqs
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class TagsModuleTest extends AnyFlatSpec with AttributesModule with Matchers {
+
+  "Empty value for tag" should "be read as empty string" in {
```
adamw, line 8: can we add some tests for non-empty values as well?

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/TagsModule.scala
```diff
@@ -46,3 +49,5 @@ trait TagsModule {
     }
   }
 }
+
+object TagsModule extends TagsModule
```
adamw, line 53: let's just use `new TagsModule {}` in tests, since this object isn't used anywhere in production code

adamw:
👍

adamw:
Thanks!

PR #515, opened 2021-07-29
Added possibilty to persist current queues configuration

### LINKED ISSUE #40: Persist queues after restart
Is there any way to persist the queues after restarting ElasticMQ? I'm not really interested in persisting the messages, just the queue itself.

bartekzylinski:
closes #40


build.sbt
```diff
@@ -80,6 +80,7 @@ val buildSettings = commonSmlBuildSettings ++ ossPublishSettings ++ Seq(
 val jodaTime = "joda-time" % "joda-time" % "2.10.10"
 val jodaConvert = "org.joda" % "joda-convert" % "2.2.1"
 val config = "com.typesafe" % "config" % "1.4.1"
+val pureConfig = "com.github.pureconfig" %% "pureconfig" % "0.16.0"
```
adamw, line 83: we'll have to switch to pure config everywhere then -> a good follow-up task

core/src/main/scala/org/elasticmq/actor/queue/QueuePersister.scala
```diff
@@ -0,0 +1,11 @@
+package org.elasticmq.actor.queue
+
+import org.elasticmq.QueueData
+
+trait QueuePersister {
+
+  def persist(queue: QueueData): Unit
```
adamw, line 7: these should probably allow async operations, that is return `Future[Unit]` (+ proper integration at the call site). 

I wonder what should happen if persisting fails ... probably propagating the exception is the best we can do

server/src/main/scala/org/elasticmq/server/QueueConfigStore.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.elasticmq.actor.queue.QueuePersister
+import org.elasticmq.server.config.ElasticMQServerConfig
+import org.elasticmq.{QueueMetadata, QueueData}
+import pureconfig.ConfigWriter
+import pureconfig.generic.auto._
+
+import java.io.PrintWriter
+import scala.collection.mutable
+
+case class QueueConfigStore(config: ElasticMQServerConfig) extends QueuePersister {
```
adamw, line 13: we definitely need a test which would verify that the serialization works

server/src/main/scala/org/elasticmq/server/QueueConfigStore.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  private val queues: mutable.Map[String, QueueData] = mutable.HashMap[String, QueueData]()
```
adamw, line 15: that's a thread-unsafe structure - we can't use it like this

server/src/main/resources/reference.conf
```diff
@@ -27,6 +27,11 @@ rest-stats {
     bind-hostname = "0.0.0.0"
 }
 
+queue-persister {
```
adamw, line 30: I think a better config name would be `queues-storage` (as a complement to `queues`), with an `enabled` flag. So sth like this:
```
queues-storage {
  enabled = false
  path = ""
}
```

core/src/main/scala/org/elasticmq/actor/queue/QueuePersister.scala
```diff
@@ -0,0 +1,11 @@
+package org.elasticmq.actor.queue
+
+import org.elasticmq.QueueData
+
+trait QueuePersister {
```
adamw, line 5: Maybe it would be easier if there was a `QueuesMetadataListener`, which would simply receive the list of current queues (the `queues` from `QueueManagerActor`) every time anything changes in any queue. We wouldn't have to maintain any state in the listener/persister implementation, and there would be only one method here

adamw: Ah ok not so easy, as we don't really have that list anywhere. Then we'll have to keep a `CHM` or sth like that in the implementation, but still I think the listener approach would be more appropriate, as the queue actors don't really have to care about persistence. Simple `QueueMetadataListener.[created | updated | remove]` would be enough?

server/src/main/scala/org/elasticmq/server/QueueConfigStore.scala
adamw: another alternative to keeping the queues state copy - upon receiving a notification, read in the current state from the file, modify, write it back

server/src/test/scala/org/elasticmq/server/config/ElasticMQServerConfigTest.scala
```diff
@@ -13,15 +13,15 @@ class ElasticMQServerConfigTest extends AnyFunSuite with Matchers with OptionVal
 
   test("load the test config") {
```
adamw, line 14: we need a test here as well, that the configuration is properly read. Additionally the persisted queue config should take precedence over config for the same queues, but defined in the config file

README.md
```diff
@@ -156,6 +156,22 @@ all messages could be either duplicated (using `copyTo` attribute) or redirected
...
+# Persisting queues configuration
+
+Queues configuration can be automatically persisted to external config file in typesafe config format.
+
+To make it work you need to pass implementation of `QueryPersister` trait while creating `SQSRestServer`. 
+To do it you can use `withQueuePersister` method from `SQSRestServerBuilder`.
+
+    SQSRestServerBuilder.withQueuePersister(QueueConfigStore(config))
+
+Additionally, you have to set `enabled` flag in `reference.conf` to true and specify `path` where the file
```
adamw, line 168: additionally? I would expect that when using the server specifying the flag would be enough. Alternatively (not additionally :) ), when using ElasticMQ in embedded mode, you don't really use the server config and can provide the persister by hand

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
```diff
@@ -27,15 +27,17 @@ class QueueManagerActor(nowProvider: NowProvider, limits: Limits) extends Replyi
             case Left(error) =>
               Left(QueueCreationError(queueData.name, error))
             case Right(_) =>
-              val actor = createQueueActor(nowProvider, queueData)
+              val actor = createQueueActor(nowProvider, queueData, queueMetadataListener)
               queues(queueData.name) = actor
+              queueMetadataListener.foreach(_ ! PersistQueue(queueData))
```
adamw, line 32: since it's a listener, maybe the naming could go along these lines: `QueueCreated`, `QueueDeleted`? This would also indicate at which point the callback is done (*after* the fact, not before)

core/src/main/scala/org/elasticmq/actor/queue/QueueActor.scala
```diff
@@ -13,7 +13,8 @@ class QueueActor(
     val initialQueueData: QueueData,
     var deadLettersActorRef: Option[ActorRef] = None,
     val copyMessagesToActorRef: Option[ActorRef] = None,
-    val moveMessagesToActorRef: Option[ActorRef] = None
+    val moveMessagesToActorRef: Option[ActorRef] = None,
+    val queueMetadataListener: Option[ActorRef]
```
adamw, line 17: default to `None` as well?

core/src/main/scala/org/elasticmq/actor/queue/QueueMetadataListener.scala
```diff
@@ -0,0 +1,7 @@
+package org.elasticmq.actor.queue
+
+import org.elasticmq.QueueData
+
+case class PersistQueue(queue: QueueData)
```
adamw, line 5: maybe group these inside an `object QueueMetadataListener`. We won't pollute global scope then. Unless other actors are designed otherwise :)

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SQSRestServerBuilder.scala
```diff
@@ -57,7 +58,8 @@ case class TheSQSRestServerBuilder(
     generateServerAddress: Boolean,
     sqsLimits: Limits,
     _awsRegion: String,
-    _awsAccountId: String
+    _awsAccountId: String,
+    queueMetadataListener: Option[ActorRef]
```
adamw, line 62: I think there should be a method `withQueueMetadataListener` to add the listener

server/src/test/scala/org/elasticmq/server/package.scala
```diff
@@ -0,0 +1,12 @@
+package org.elasticmq
+
+import scala.io.Source
+
+package object server {
+  private[server] def load[T](base: Class[T], fileName: String): String = {
```
adamw, line 6: `loadResource`?

server/src/main/scala/org/elasticmq/server/QueueMetadataListener.scala
```diff
@@ -0,0 +1,27 @@
+package org.elasticmq.server
+
+import akka.actor.Actor
+import org.elasticmq.QueueData
+import org.elasticmq.actor.queue.{PersistQueue, RemoveQueue, UpdateQueueMetadata}
+
+import scala.collection.mutable
+
+class QueueMetadataListener(storagePath: String) extends Actor {
```
adamw, line 9: The name here should be `QueueConfigStore` or similar, since we're implementing the listener for a specific use-case - not a general-purpose listener

server/src/main/scala/org/elasticmq/server/QueueMetadataListener.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  def receive: Receive = {
+    case PersistQueue(queue) =>
+      queues.put(queue.name, queue)
+      queuePersister.saveToConfigFile(queues.values.toList)
+    case RemoveQueue(queueName) =>
+      queues.remove(queueName)
+      queuePersister.saveToConfigFile(queues.values.toList)
+    case UpdateQueueMetadata(queue) =>
+      queues.remove(queue.name)
```
adamw, line 23: doesn't put override?

server/src/main/scala/org/elasticmq/server/QueuePersister.scala
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  private def serializeQueueMetadata(queueNameToMetadata: (String, QueueMetadata)): String = {
+    val options = ConfigRenderOptions.defaults().setOriginComments(false).setJson(false).setFormatted(false)
+    val str = ConfigWriter[QueueMetadata].to(queueNameToMetadata._2).render(options)
+    s" ${queueNameToMetadata._1} {$str}\n"
+  }
+
+  private def toQueueConfig(queue: QueueData): QueueMetadata =
+    QueueMetadata(
```
adamw, line 41: is this the only place where we are using `QueueMetadata`? If so, let's move this class and make it private here

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
```diff
@@ -68,47 +71,75 @@ class ElasticMQServerConfig(config: Config) extends Logging {
...
+  }
+
+  val createBaseQueues: Seq[CreateQueue] = {
+    val baseQueues: mutable.Map[String, ConfigValue] = config
+      .getObject("queues")
+      .asScala
+    createQueuesFromConfig(baseQueues)
+  }
+
+  private val persistedQueuesConfig: Option[Config] =
```
adamw, line 91: let's keep a single name for the stored queues. Currently we have:
* queue storage
* persisted queues
* backup file

Let's keep one of: storage, persistence, backup

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
```diff
@@ -68,47 +71,75 @@ class ElasticMQServerConfig(config: Config) extends Logging {
...
+      case Some(file) =>
+        val queuesConfig: mutable.Map[String, ConfigValue] = file
+          .getObject("queues")
+          .asScala
+        createQueuesFromConfig(queuesConfig)
+      case None => Nil
+    }
 
-  val createQueues: List[CreateQueue] = {
+  def createQueuesFromConfig(queuesConfig: mutable.Map[String, ConfigValue]): List[CreateQueue] = {
```
adamw, line 109: if this doesn't depend on other internal state of `ElasticMQServerConfig`, maybe let's move it to the companion object so that we're sure it's a "stateless" function?

server/src/main/scala/org/elasticmq/server/config/ElasticMQServerConfig.scala
```diff
@@ -68,47 +71,75 @@ class ElasticMQServerConfig(config: Config) extends Logging {
...
+
+  private val persistedQueuesConfig: Option[Config] =
+    if (queuesStorageEnabled)
+      Some(
+        ConfigFactory
+          .parseFile(new File(queuesStoragePath))
+      )
+    else None
+
+  def createPersistedQueues(backupFile: Option[Config] = persistedQueuesConfig): List[CreateQueue] =
```
adamw, line 99: the flow here isn't easy to understand. First, maybe let's switch from `createPersistenQueues` to `readPersistedQueues` (and same above). I know the name used to be `createQueues` but that was just a copy of the result type, probably not the most readable name as well :)

Secondly, the main entry point here is `readQueuesToLoad` which in production will always be called with the `createPersistedQueues` parameter - so maybe let's make it the default?

server/src/test/scala/org/elasticmq/server/QueuePersisterTest.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.elasticmq.server.config.ElasticMQServerConfig
+import org.elasticmq.{DeadLettersQueueData, MillisVisibilityTimeout, QueueData}
+import org.joda.time.{DateTime, Duration}
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class QueuePersisterTest extends AnyFunSuite with Matchers {
+
+  test("should parse queue data") {
+    val conf = new ElasticMQServerConfig(ConfigFactory.load("test"))
```
adamw, line 13: why are we loading config in a `QueuePersister` unit-test?

server/src/test/resources/backup.conf
```diff
@@ -0,0 +1,3 @@
+queues {
```
adamw, line 1: a more indicative name of this file would be good :) it's a test file, right? not a backup of something?

core/src/main/scala/org/elasticmq/actor/queue/QueueMetadataListener.scala
bartekzylinski: From what I see for `QueueManagerActor` and `QueueActor` messages are not grouped in any way.

adamw: Ok, maybe a trait which would be extended by all messages accepted by the actor? So that the protocol is clear

server/src/test/scala/org/elasticmq/server/config/ElasticMQServerConfigTest.scala
```diff
@@ -32,13 +33,59 @@ class ElasticMQServerConfigTest extends AnyFunSuite with Matchers with OptionVal
...
+      Some(0L),
+      Some(DeadLettersQueue("dead-2", 4)),
+      isFifo = false,
+      hasContentBasedDeduplication = true,
+      Some("copyTo"),
+      Some("messageTo"),
+      Map("tag12Key" -> "tag12Value")
+    )
+    persistedQueues.length shouldBe 3
+    persistedQueues.head shouldBe expectedQueue
```
adamw, line 89: this is the modified queue? shouldn't we check the other ones in the list as well?

adamw: would be helpful to somehow write in the test which queues are from the conf, which from the persistence file, and how they are modified

Issue #522, opened 2021-08-25
SequenceNumber missing from Message Attributes (SQS fifo)


terozio:
When doing a receive-message for a fifo queue message attributes should contain attribute called SequenceNumber (to indicate the ordering within a message group in a fifo queue). https://docs.aws.amazon.com/cli/latest/reference/sqs/receive-message.html#output

Steps to replicate:
1. `docker run -p 9324:9324 -p 9325:9325 softwaremill/elasticmq-native:1.2.0`
2. `aws sqs create-queue --queue-name test-queue.fifo --attributes FifoQueue=true --endpoint-url http://localhost:9324"`
3. `aws sqs send-message --queue-url http://localhost:9324/queue/test-queue.fifo --message-body "msg body" --message-group-id "my-group" --message-deduplication-id "1234" --endpoint-url http://localhost:9324`
4. `aws sqs receive-message --queue-url http://localhost:9324/queue/test-queue.fifo --attribute-names All --endpoint-url http://localhost:9324`
5. Observe that the response does not contain a field called `SequenceNumber` inside attributes

Our usecase is testing Lambda functions locally, we use SequenceNumber for idempotency purposes when the lamba stores the last message that has been successfully processed.

Not sure if this is a bug or that the fifo functionality is just partially implemented, didn't notice any explicit statement on the documentation.

pierscin:
It appears to be an issue in elastic.

According to [sqs docs](https://docs.aws.amazon.com/cli/latest/reference/sqs/send-message.html#output):

`SequenceNumber -> (string)`

This parameter applies only to FIFO (first-in-first-out) queues.
The large, non-consecutive number **that Amazon SQS assigns to each message**.
The length of SequenceNumber is 128 bits. SequenceNumber continues to increase for a particular MessageGroupId.

adamw:
Fixed in 1.2.1

peebles:
I am still not seeing this attribute, nor am I seeing the FifoQueue attribute.

I am using ELQ version 1.2.2:
21:28:26.792 [main] INFO  org.elasticmq.server.Main$ - Starting ElasticMQ server (1.2.2) ...

My queue looks like:
$ aws --endpoint-url http://localhost:9324 sqs get-queue-attributes --queue-url "http://localhost:9324/queue/zc-ingest-local.fifo" --attribute-names All
{
    "Attributes": {
        "ApproximateNumberOfMessagesNotVisible": "0", 
[...]

PR #527, opened 2021-09-08
Add sequence number to FIFO queue

pierscin:
According to [Message docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html), there should be additional attribute in `ReceiveMessage` - `SequenceNumber`. Other than that `SequenceNumber` should be returned in `SendMessage` result as per [this doc](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html).

Number should be increasing per GroupId.

Few things I've noticed on AWS:
- SequenceNumber is increasing globally, but in docs it is specified that it should increase at least per GroupId.
- SequenceNumber passed as an attribute in endMessage is ignored when receiving the same message

Connected with issue #522

### DIFF
--- a/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
@@ -83,8 +83,7 @@ case class InternalMessage(
 }
 
 object InternalMessage {
-  def from(newMessageData: NewMessageData, queueData: QueueData, sequenceNumber: BigInt): InternalMessage = {
-    val attributes = updateSequenceNumberAttribute(newMessageData, queueData, sequenceNumber)
+  def from(newMessageData: NewMessageData, queueData: QueueData): InternalMessage = {
 
     val now = System.currentTimeMillis()
     new InternalMessage(
@@ -92,7 +91,7 @@ object InternalMessage {
       mutable.Buffer.empty,
       newMessageData.nextDelivery.toMillis(now, queueData.delay.getMillis).millis,
       newMessageData.content,
-      attributes,
+      newMessageData.messageAttributes,
       new DateTime(),
       newMessageData.orderIndex,
       NeverReceived,
@@ -105,16 +104,4 @@ object InternalMessage {
   }
 
   private def generateId() = MessageId(UUID.randomUUID().toString)
-
-  private def updateSequenceNumberAttribute(
-      newMessageData: NewMessageData,
-      queueData: QueueData,
-      sequenceNumber: BigInt
-  ) = {
-    if (!queueData.isFifo) {
-      (newMessageData.messageAttributes - "SequenceNumber")
-    } else {
-      newMessageData.messageAttributes.updated("SequenceNumber", StringMessageAttribute(sequenceNumber.toString()))
-    }
-  }
 }
--- a/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
@@ -84,11 +84,11 @@ case class InternalMessage(
 
 object InternalMessage {
 
-  def from(newMessageData: NewMessageData, queueData: QueueData, n: BigInt): InternalMessage = {
+  def from(newMessageData: NewMessageData, queueData: QueueData, sequenceNumber: BigInt): InternalMessage = {
     val attributes = newMessageData
       .messageAttributes
       .updatedWith("SequenceNumber")(
-      _ => if (queueData.isFifo) Some(StringMessageAttribute(n.toString())) else None
+      _ => if (queueData.isFifo) Some(StringMessageAttribute(sequenceNumber.toString())) else None
     )
 
     val now = System.currentTimeMillis()
--- a/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
@@ -84,14 +84,20 @@ case class InternalMessage(
 
 object InternalMessage {
 
-  def from(newMessageData: NewMessageData, queueData: QueueData): InternalMessage = {
+  def from(newMessageData: NewMessageData, queueData: QueueData, n: BigInt): InternalMessage = {
+    val attributes = newMessageData
+      .messageAttributes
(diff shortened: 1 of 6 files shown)


rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/AmazonJavaSdkTestSuite.scala
```diff
@@ -834,6 +834,77 @@ class AmazonJavaSdkTestSuite extends SqsClientServerCommunication with Matchers
...
+
+    val seqNum21 = BigInt(messages(2).getMessageAttributes.asScala("SequenceNumber").getStringValue)
+    val seqNum22 = BigInt(messages(3).getMessageAttributes.asScala("SequenceNumber").getStringValue)
+
+    groupId1 shouldNot equal(groupId2)
+
+    seqNum11 should be < seqNum12
+    seqNum21 should be < seqNum22
+
+    seqNum11 should equal(seqNum21)
```
adamw, line 904: they *can* be equal, or *must*? :)

pierscin: I had a problem with this too. "**can**" is the answer.

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SendMessageDirectives.scala
```diff
@@ -35,6 +35,12 @@ trait SendMessageDirectives { this: ElasticMQDirectives with SQSLimitsModule =>
                   {messageAttributeDigest.map(d => <MD5OfMessageAttributes>{d}</MD5OfMessageAttributes>).getOrElse(())}
                   <MD5OfMessageBody>{digest}</MD5OfMessageBody>
                   <MessageId>{message.id.id}</MessageId>
+                  {message.messageAttributes.get("SequenceNumber").flatMap(sn => {
```
adamw, line 38: shouldn't we use `SequenceNumberAttribute` here?

core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
```diff
@@ -83,15 +83,16 @@ case class InternalMessage(
 }
 
 object InternalMessage {
+  def from(newMessageData: NewMessageData, queueData: QueueData, sequenceNumber: BigInt): InternalMessage = {
+    val attributes = updateSequenceNumberAttribute(newMessageData, queueData, sequenceNumber)
```
adamw, line 87: doesn't this assign a sequence number to every message, instead of only FIFO messages?

core/src/main/scala/org/elasticmq/actor/queue/QueueActorStorage.scala
```diff
@@ -13,8 +13,15 @@ trait QueueActorStorage {
...
+    val next = sequenceNumber
+    sequenceNumber = sequenceNumber + 1
+    next
+  }
+
   var queueData: QueueData = initialQueueData
   var messageQueue: MessageQueue = MessageQueue(queueData.isFifo)
   var fifoMessagesHistory: FifoDeduplicationIdsHistory = FifoDeduplicationIdsHistory.newHistory()
   val receiveRequestAttemptCache = new ReceiveRequestAttemptCache
+  var sequenceNumber = BigInt(0)
```
adamw, line 26: wouldn't a `Long` be enough? :)

core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
pierscin: No it doesn't. It creates number for every message in queue but doesn't assign it. If user sent `SequenceNumber` in attribute it will delete it.

core/src/main/scala/org/elasticmq/actor/queue/QueueActorStorage.scala
pierscin: It probably would, but in specs it says that SequenceNumber should be 128-bit.

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SendMessageDirectives.scala
pierscin: :+1:

core/src/main/scala/org/elasticmq/actor/queue/operations/SendMessageOp.scala
```diff
@@ -37,7 +37,7 @@ trait SendMessageOp extends Logging {
   }
 
   private def addMessage(message: NewMessageData) = {
-    val internalMessage = InternalMessage.from(message, queueData)
+    val internalMessage = InternalMessage.from(message, queueData, nextSequenceNumber())
```
adamw, line 40: wouldn't it make more sense to adapt the message attributes here, and add the sequence number, before converting to an `InternalMessage`? That way we would avoid passing things around a bit

core/src/main/scala/org/elasticmq/actor/queue/QueueActorStorage.scala
adamw: Hm, if we always use a `Long`, that requirement will be satisfied :)

Also, maybe this var can be private?

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/SendMessageDirectives.scala
adamw: ok, you're right :)

core/src/main/scala/org/elasticmq/actor/queue/InternalMessage.scala
pierscin: moving method fixed this

core/src/main/scala/org/elasticmq/actor/queue/operations/SendMessageOp.scala
pierscin: 👍

PR #605, opened 2021-12-25
CI refactor

Pask423:
Firstly number versions of Scala in `build.sbt` were not in line with CI's versions.
Additionally `mergify` was using hardcoded names of jobs with old versions.
I removed `matrix` from workflow definition and split it to two separate jobs.
When updating Scala versions in `build.sbt` we should remember to update one in GH actions too.


micossow (approved):
looks good :)

adamw (approved):

PR #616, opened 2022-01-03
upgrade sml dependencies to 2.x

### LINKED ISSUE #596: Move release to be run using gh actions
Currently this is a local process. Moving this to a CI job will make things easier to run without the need for local setup.

This involves updating the sbt-softwaremill plugins to 2.x.

micossow:
fixes #596


adamw:
Looks good :) To test if this works, maybe let's try merging and doing a couple of test releases (either minor or e.g `1.3.4-RCn`). For the testing I would swap the ordering of the steps though, so that the jar artifact upload is at the end (as it's irreversible), and publishing the release notes last (as it notifies people that there's a new release). This way if anything goes wrong, we'll be able to revert or at least fix without notifying too many people :)

.github/workflows/ci.yml
```diff
@@ -65,4 +66,114 @@ jobs:
...
+      - name: Compile
+        run: sbt compile
+
+      - name: Build jvm docker image
+        run: sbt "project server; assembly; Docker / publishLocal"
+
+      - name: Build native docker image
+        run: sbt "project nativeServer; assembly; GraalVMNativeImage / packageBin; Docker / publishLocal"
+
+  publish:
```
adamw, line 105: shouldn't we have a dependency here so that this is only run if ci-* jobs pass?

micossow: it's temporarily removed to speed up testing

adamw:
Looks great - thanks! :)

PR #619, opened 2022-01-05
add e2e test with docker images resolving issue #582

### LINKED ISSUE #582: Messages are not moved to DLQ
I'm using ElasticMQ (via docker image v1.3.3) and can't get the DLQ to work.

This is my `elasticmq.conf`:

```
include classpath("application.conf")

node-address {
   protocol = http
   host = localhost
   port = 9324
   context-path = ""
}

rest-sqs {
   enabled = true
   bind-port = 9324
   bind-hostname = "0.0.0.0"
   // Possible values: relaxed, strict
   sqs-limits = strict
}

rest-stats {
  enabled = true
  bind-port = 9325
  bind-hostname = "0.0.0.0"
}

queues {
   main {
     defaultVisibilityTimeout = 10 seconds
     delay = 2 seconds
     deadLettersQueue {
       name = "retry"
       maxReceiveCount = 1
     }
   }
   retry {
     defaultVisibilityTimeout = 10 seconds
     delay = 2 seconds
   }
   deadletter {
     defaultVisibilityTimeout = 10 seconds
     delay = 2 seconds
   }
}
```

I'm sending a message like so (using the AWS CLI):
```
aws --endpoint-url http://localhost:9324 sqs send-message --queue-url http://localhost:9324/queue/main --message-body "Hello, queue"
```

And receiving it like so:
```
aws --endpoint-url http://localhost:9324 sqs receive-message --queue-url http://localhost:9324/queue/main --wait-time-seconds 10
```

I'm not deleting the message from the queue but the message is being deleted and not being moved to the DLQ (i.e., the retry queue). I'm also trying to receive the message with Java code and getting the same result.

Why is that?

szymonneo:
closes #582

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,6 +25,7 @@ val buildSettings = commonSmlBuildSettings ++ ossPublishSettings ++ Seq(
   scalaVersion := v2_13,
   crossScalaVersions := Seq(v2_13, v2_12),
   libraryDependencies += "org.scala-lang.modules" %% "scala-xml" % "1.3.0",
+  libraryDependencies += "com.dimafeng" %% "testcontainers-scala-scalatest" % "0.39.12" % "test",
   dependencyOverrides := akka25Overrides,
   parallelExecution := false,
   sonatypeProfileName := "org.elasticmq",
--- a/rest/rest-sqs-testing-amazon-java-sdk/src/test/resources/custom.conf
+++ b/rest/rest-sqs-testing-amazon-java-sdk/src/test/resources/custom.conf
@@ -1,0 +1,44 @@
+include classpath("application.conf")
+
+node-address {
+   protocol = http
+   host = localhost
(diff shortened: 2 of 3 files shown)


adamw:
Thanks :)

So was there any problem in the code, or is it just a test verifying that things work as expected?

Second question - is there anything special in this test that makes it impossible to write using the current approach (starting an embedded server & using the java sdk)?

micossow:
My requirement wasn't clear enough and that's why this PR has been created. We wanted a simple repro script, that @szymonneo eventually did (see https://github.com/softwaremill/elasticmq/issues/582#issuecomment-1005578688). And we don't want to merge this - it doesn't test any piece of code. Besides, this scenario is already tested in SDK test suite.

Anyway, thanks @szymonneo for a good work, maybe we can use this in future.

PR #654, opened 2022-02-18
Refactor MessageQueue and fix perf issue

### LINKED ISSUE #646: Low read throughput when queue has retained a lot of messages.
We've been doing some nice SQS tests thanks to elasticMQ and were planning to use it for some load testing to understand throughput on our code, and while write results were very nice and definitely are on par to what is described on the README, read throughput was very low at an high message count, and it seemed to degrade the more messages were retained in the queue.

Here's our scenario for more context.

Queue configuration: 
```
queues {
  queue-test {
    defaultVisibilityTimeout = 10 seconds
    delay = 0 seconds
    fifo = true
    contentBasedDeduplication = false
    tags {
      tag1 = "tagged1"
      tag2 = "tagged2"
    }
  }
}
```

We wrote around 216000 messages of varying size (don't have an exact number, but not bigger than a few kb)  without consuming it, to this queue without any discernable delay, once they were all within the queue, we started to consume them with `spring-cloud-aws`, and we were only receiving around 3~ messages/second, testing different implementation and approacches (e.g, multiple consumers, other libraries) yielded the same rate. 

After making sure the bottleneck was not in the implementation side, we did tests with different messsage sizes and it seems that the more messages there is, the slower it would become, at around 80k messages the rate was 15 messages/second, at 150k it was 5 messages/second, for comparison sake, writing and reading at the same time so the queue would not have any messages retained gave us rates of 100-150 messages/second, 

As far as I tested fifo vs standard queues reacted the same, docker or the jar standalone (v1.3.3) made no difference, and we didn't see any obvious problem monitoring the jar through JMX 

Is this a known limitation and out of the scope of elasticMQ or a legitimate bug? it has been very useful for our automatic and manual tests so it was sad to see it would choke while doing load tests. 

Thanks for the help!

micossow:
fixes #646

### DIFF
--- a/core/src/main/scala/org/elasticmq/actor/queue/MessageQueue.scala
+++ b/core/src/main/scala/org/elasticmq/actor/queue/MessageQueue.scala
@@ -13,12 +13,21 @@ sealed trait MessageQueue {
     */
   def +=(message: InternalMessage): Unit
 
-  /** Get the messages indexed by their unique id
+  /** Get the message by its unique id
+    *
+    * @param id
+    *   The message id
+    * @return
+    *   The message
+    */
+  def getById(id: String): Option[InternalMessage]
+
+  /** Get all messages in queue
     *
     * @return
-    *   The messages indexed by their id
+    *   All messages in queue
     */
-  def byId: Map[String, InternalMessage]
+  def all: Iterable[InternalMessage]
 
   /** Drop all messages on the queue
     */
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/elasticmq/actor/queue/MessageQueue.scala
```diff
@@ -126,7 +135,9 @@ object MessageQueue {
       messageQueue += message
     }
 
-    override def byId: Map[String, InternalMessage] = messagesById.toMap
```
micossow, line 129: `messagesById.toMap` triggered unnecessary copy which increased the complexity to O(n)

adamw (approved):

Issue #673, opened 2022-04-07
Messages are not moved to DLQ when the DLQ name contains underscores


Stanleyyyan:
We observed some weird behaviors with the DLQ's name containing underscores. When the ReceiveCount for a message exceeds the maxReceiveCount for a queue, the message is deleted from the main queue (the consumer doesn't remove the message from the main queue) and not moved to DLQ as expected.

To reproduce the issue, try configuring a DLQ with the name of "dlq_some_queue".

Pask423:
Just one question, how do you create your queues?

micossow:
@Stanleyyyan can you check if the issue still occurs with `1.3.7`?

Stanleyyyan:
@Pask423, we ran the stand-alone via docker ([link](https://github.com/softwaremill/elasticmq#elasticmq-via-docker))

@micossow, Yes, it's working now!  Thanks for fixing that!

PR #677, opened 2022-04-19
Queue ordering fix

Pask423:
After merging and releasing ask people from #673 if this solved their problem


micossow (approved):
Thanks :D

PR #678, opened 2022-04-21
Added readme line about how to fix mounting problem on windows

### LINKED ISSUE #301: Mount directories instead of files in Docker
Mounting files is only supported on Linux and not portable to other OSs.
This makes it unnecessarily difficult to use this docker image on Windows and other operating systems.
For details take a look at https://github.com/moby/moby/issues/30555.

Pask423:
closes #301


micossow (approved):
Thanks :)

PR #703, opened 2022-07-15
remove scala versions from gh workflow

### LINKED ISSUE #695: hardcoded scala version in ci.yml
Keep specific scala versions in build.sbt only. 

Hardcoding scala versions in ci.yml cause problems with automatic version upgrade. 

See for example this error: https://github.com/softwaremill/elasticmq/runs/6976031061?check_suite_focus=true
In the above example the scala-steward cannot push a PR with changes in ci.yml because it lacks permissions to modify workflows. To open such permissions seems unnecessary.

micossow:
fixes #695

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,8 +1,8 @@
 import com.amazonaws.services.s3.model.PutObjectResult
-import com.softwaremill.SbtSoftwareMillCommon.commonSmlBuildSettings
 import com.softwaremill.Publish.ossPublishSettings
+import com.softwaremill.SbtSoftwareMillCommon.commonSmlBuildSettings
 import com.typesafe.sbt.packager.docker.{Cmd, ExecCmd}
-import sbt.Keys.{credentials, javaOptions}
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -11,6 +11,10 @@ import scala.sys.process.Process
 val v2_12 = "2.12.15"
 val v2_13 = "2.13.8"
 
+lazy val resolvedScalaVersion = sys.env.get("SCALA_MAJOR_VERSION") match {
+  case Some(majorVersion) => if (majorVersion == "2.12") v2_12 else v2_13
```
mkrzemien, line 15: Minor: what about the following?
```
case Some(majorVersion) if majorVersion == "2.12" => v2_12
case Some(majorVersion) if majorVersion == "2.13" => v2_13
case _ => 2_13
```

mkrzemien (approved):

build.sbt
adamw: Or `case Some("2.12")` :)

PR #759, opened 2022-11-08
Restore java 8 support

micossow:


build.sbt
```diff
@@ -89,13 +89,13 @@ val s3Upload = TaskKey[PutObjectResult]("s3-upload", "Uploads files to an S3 buc
...
-  .settings(name := "elasticmq-root", publishArtifact := false)
-  .aggregate(commonTest, core, rest, persistence, server, nativeServer, ui)
+  .settings(name := "elasticmq-root", publish / skip := true)
+  .aggregate(commonTest, core, rest, persistence, server, ui)
 
 lazy val commonTest: Project = (project in file("common-test"))
   .settings(buildSettings)
   .settings(name := "elasticmq-common-test")
-  .settings(Seq(libraryDependencies ++= Seq(scalatest, awaitility, logback), publishArtifact := false))
+  .settings(Seq(libraryDependencies ++= Seq(scalatest, awaitility, logback), publish / skip := true))
```
adamw, line 98: shouldn't this remain in aggregates?

adamw (approved):

build.sbt
```diff
@@ -89,13 +89,13 @@ val s3Upload = TaskKey[PutObjectResult]("s3-upload", "Uploads files to an S3 buc
 lazy val root: Project = (project in file("."))
   .enablePlugins(GitVersioning)
   .settings(buildSettings)
-  .settings(name := "elasticmq-root", publishArtifact := false)
-  .aggregate(commonTest, core, rest, persistence, server, nativeServer, ui)
+  .settings(name := "elasticmq-root", publish / skip := true)
+  .aggregate(commonTest, core, rest, persistence, server, ui)
```
adamw, line 93: shouldn't `nativeServer` remain in aggregates?

micossow: Here's the tricky part, we want to build the main jar using java 8, but native-server requires java 11, so it's built separately. Besides, there's nothing to build/test there within the main pipeline so it can be safely removed from root. It's still used when building native image ofc.

adamw: aah ok - let's add a comment why it's not in the aggregates, it's non-obvious :)

PR #781, opened 2023-01-09
Update Alpine base docker image.

daniel-cowgill-rh:
The current version of alpine that is used to build the docker image on master is 3.11.

That has image has a critical zlib vulnerability. https://dso.docker.com/cve/CVE-2022-37434

This updates to 3.17


micossow:
Thanks for the contribution :)

PR #817, opened 2023-05-19
issue 812 add support for json

katlasik:


rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/CreateQueueDirectives.scala
```diff
@@ -115,6 +117,21 @@ trait CreateQueueDirectives {
   }
 }
 
+case class CreateQueueRequest(QueueName: String, Attributes: Option[Map[String, String]], tags: Option[Map[String, String]])
+
+object CreateQueueRequest {
+  implicit val format: RootJsonFormat[CreateQueueRequest] = jsonFormat3(CreateQueueRequest.apply)
+
+  implicit val fpr: FlatParamsReader[CreateQueueRequest] = new FlatParamsReader[CreateQueueRequest] {
```
flsh86, line 125: Maybe we change a variable name here? It's hard to say what it is without looking at the type + maybe we also change `format` so it's more verbose.

Maybe something like
`requestJsonFormat` or `requestJsonFormatter`
and
`requestParamsReader`
wdyt?

rest/rest-sqs/src/main/scala/org/elasticmq/rest/sqs/model/RequestPayload.scala
```diff
@@ -0,0 +1,36 @@
...
+import akka.http.scaladsl.server.Directive0
+import akka.http.scaladsl.server.Directives._
+import org.elasticmq.rest.sqs.{Action, FlatParamsReader}
+import org.elasticmq.rest.sqs.model.RequestPayload.QueryParams
+import spray.json.{JsObject, RootJsonFormat}
+
+sealed trait RequestPayload {
+  def action(requiredAction: Action.Value): Directive0
+
+  def maybeAction: Option[String]
```
flsh86, line 12: The method name doesn't sound right to me. It doesn't really say what's the meaning of that method even if the implementation is passed to class implementing that Trait.

It's used in two places:
```
override def maybeAction: Option[String] = params.get("Action") //Map[String, String]
override def maybeAction: Option[String] = Some(action)
```

In first we're `getting` an `Action` and in second we're wrapping it but the end result is us retrieving an optinal value of Action, so maybe `retrieveAction` might be a better name here?

PR #832, opened 2023-06-15
Don't fail if datatype contains additional characters

katlasik:


katlasik:
Closes #829

adamw (approved):
Nice, there's even a Spring test :)

Issue #724, opened 2022-09-01
add native arm image


realies:
currently, it seems like elasticmq-native runs through qemu-x86_64 on arm, could this be made into a native image? it would really benefit laptop development and battery life, since it currently takes about 10-20% cpu with a light polling client

micossow:
Continuation of #473

realies:
thanks @micossow, as I mentioned there, there's darwin aarch64 support over at https://github.com/graalvm/graalvm-ce-dev-builds/releases now, could m1 arm native elasticmq-native images be a thing with it?

micossow:
It should be possible to build such image now, see: https://github.com/sbt/sbt-native-packager/issues/1503
Some changes in build.sbt are probably also necessary.

fvasco:
I tested the image `softwaremill/elasticmq-native:1.3.14` on a Graviton AWS server, but it does not work

~~~
# docker run softwaremill/elasticmq-native:1.3.14
Unable to find image 'softwaremill/elasticmq-native:1.3.14' locally
1.3.14: Pulling from softwaremill/elasticmq-native
Digest: sha256:c251751b5d668033c69494ff37a938f977397ad3c806085e53f269bedb2c3ead
Status: Downloaded newer image for softwaremill/elasticmq-native:1.3.14
[FATAL tini (7)] exec /opt/docker/bin/elasticmq-native-server failed: Exec format error
~~~

The executable in the image is in amd64 format

~~~
file opt/docker/bin/elasticmq-native-server
opt/docker/bin/elasticmq-native-server: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, for GNU/Linux 3.2.0, BuildID[sha1]=2ad8fd7fe117da1b01aea2b469d54497c42743d3, with debug_info, not stripped, too many notes (256)
~~~

Other files look for aarch64 architecture.

kderholtvisma:
Will this be fixed? or is there a quick way of building this from the repo manually?

micossow:
If you need docker image locally, you can run `sbt "project nativeServer; clean; assembly; graalvm-native-image:packageBin; docker:publishLocal"` on your AMR64 system.

To be able to create native docker images as a part of GH workflow, we need either:
* GH Workflow supporting ARM64 systems
* or cross compilation support in GraalVM https://github.com/oracle/graal/issues/407

diegobfernandez:
[...]

PR #892, opened 2023-10-27
Native image cross arch docker build support

micossow:


micossow:
closes #724

adamw:
That's hard to review as I never really implemented sth like that, I can only suggest merging & doing an RC release to verify if this works, otherwise iterate on a fix :). My only question I suppose would be, where's the `buildx` invocation?

micossow:
> That's hard to review as I never really implemented sth like that, I can only suggest merging & doing an RC release to verify if this works, otherwise iterate on a fix :). My only question I suppose would be, where's the `buildx` invocation?

It has been verified with https://github.com/softwaremill/elasticmq/releases/tag/v1.4.5-RC2
`buildx` is configured in `dockerBuildxSettings`

adamw:
@micossow Well if it's verified, then ship it :)

Issue #896, opened 2023-11-01
Automatic creation of queue does not work


Kytozz:
I updated elasticmq version from 1.4.5 to 1.4.6 and now automatic queue generation is not working.
Any ideas on this issue?

```yaml
# docker-compose.yml
  elasticmq:
    build:
      context: .
      dockerfile: Dockerfile-elasticmq
    image: elasticmq:1.0.0
    ports:
      - "9324:9324"
      - "9325:9325"
    volumes:
      - ./config/elasticmq/custom.conf:/opt/elasticmq.conf
```

``` dockerfile
# Dockerfile-elasticmq
FROM softwaremill/elasticmq-native:1.4.6
```

```
# config/elasticmq/custom.conf
include classpath("application.conf")

queues {
  "default-dead-letters" {
    defaultVisibilityTimeout = 600 seconds
    delay = 0 seconds
    receiveMessageWait = 0 seconds
  }
  "default" {
    defaultVisibilityTimeout = 600 seconds
    delay = 0 seconds
    receiveMessageWait = 0 seconds
    deadLettersQueue {
      name = "default-dead-letters.fifo"
      maxReceiveCount = 100
    }
  }
}
```

![Screenshot 2023-11-01 at 18 07 28](https://github.com/softwaremill/elasticmq/assets/106790619/05ed0773-f0d6-48c0-8d8e-7d8d1043d2a4)

micossow:
Thanks for reporting the issue. It should be fixed in `v1.4.7`

PR #897, opened 2023-11-01
fix configuration files paths in native docker image

micossow:


micossow:
fixes #897

micossow:
fixes #896

PR #904, opened 2023-11-15
Update README.md

WNC-Travis:
I was stumped for a while because my queues were not being created when I named them `queue1.fifo` in the config file.  This is because the `.` is used to delineate parameter names.

Feel free to edit this if there is a way to actually escape the `.`, but I could not find that solution.


WNC-Travis:
Found this: `single underscore(_) is converted into a dot(.)`

PR #953, opened 2024-01-22
fix: The JSON protocol does not handle empty lists, so use the Option

jkugiya:
https://github.com/localstack/localstack/pull/9710/files#diff-afa491974d1d7cb931555d6bfd9bfbf8e4f977f1528242ba8732413a848c15b7R251-R252

Similar fix to a recent fix in Localstack: In JSON Protocol, `BatchResponse` does not seem to return "failed" as a field on successful request. If it did, `SendMessageBatchResponse#hasFailed` would not work correctly. (https://github.com/aws/aws-sdk-java-v2/issues/4759)


micossow:
Thanks for contribution!

PR #987, opened 2024-04-05
Implement DLQ redrive actions

micossow:


core/src/main/scala/org/elasticmq/actor/queue/operations/MoveMessagesAsyncOps.scala
```diff
@@ -0,0 +1,159 @@
...
+            sourceArn,
+            startedTimestamp,
+            taskHandle
+          ) =>
+        logger.debug("Trying to move a single message to {} ({} messages left)", destinationQueue, messageQueue.size)
+        messageQueue.pop match {
+          case Some(internalMessage) =>
+            messageMoveTaskState = mmInProgress.copy(numberOfMessagesMoved = numberOfMessagesMoved + 1)
+            destinationQueue ! SendMessage(internalMessage.toNewMessageData)
+            maxNumberOfMessagesPerSecond match {
```
kciesielski, line 90: **question:** Do we need to implement the throttling? I think the `maxNumberOfMessagesPerSecond` may be relevant for real backends to avoid performance issues, but is it a case for ElasticMQ? Maybe users who use it as a stub would prefer to have this done immediately?
Edit: Ok, there may be a use case where users want to test their cancellation logic, so they would use this parameter in tests to emulate a slow cancellable process, like you did, right?

micossow: Exactly, I'm using this feature in integration tests as well. Beside that, it's quite easy to emulate it with Akka.

core/src/main/scala/org/elasticmq/ElasticMQError.scala
```diff
@@ -1,8 +1,9 @@
 package org.elasticmq
+import org.elasticmq.msg.MessageMoveTaskHandle
 
 trait ElasticMQError {
   val queueName: String
-  val code: String
+  val code: String // TODO: code should be handled in rest-sqs module
```
kciesielski, line 6: **question:** Does this TODO mean that there shouldn't be `code` here?

core/src/main/scala/org/elasticmq/actor/QueueManagerActor.scala
```diff
@@ -7,18 +7,23 @@ import org.elasticmq.actor.reply._
...
+
   type M[X] = QueueManagerMsg[X]
   val ev: ClassTag[QueueManagerMsg[Unit]] = classTag[M[Unit]]
 
-  case class ActorWithQueueData(actorRef: ActorRef, queueData: QueueData)
-  private val queues = collection.mutable.HashMap[String, ActorWithQueueData]()
+  val queues: mutable.Map[MessageMoveTaskHandle, ActorWithQueueData] = mutable.HashMap[String, ActorWithQueueData]()
 
-  def receiveAndReply[T](msg: QueueManagerMsg[T]): ReplyAction[T] =
+  // TODO: create *Ops class like in QueueActor
```
kciesielski, line 24: **question:** Is this supposed to be implemented in a subsequent PR with refactoring, to avoid extending the scope too much?

micossow: Yes, this is for future me

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/MessageMoveTaskTest.scala
```diff
@@ -0,0 +1,208 @@
...
+
+    (queue, dlq)
+  }
+
+  private def receiveAllMessagesTwice(queue: QueueUrl): Unit = {
+    for (_ <- 0 until NumMessages) {
+      testClient.receiveMessage(queue)
+    }
+
+    Thread.sleep(1500)
```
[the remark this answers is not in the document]
micossow: I don't think we could use this even if the core was implemented in Akka typed, because we only do black box testing via AWS SDK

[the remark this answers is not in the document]
kciesielski: I think one could inject a `ManualTimer` somehow during test initialization, and then perhaps mix SDK calls with `timer.timePasses` calls, but let's leave this as it is.

kciesielski (approved):

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/MessageMoveTaskTest.scala
```diff
@@ -0,0 +1,208 @@
...
+  test("should run message move task with max number of messages per second") {
+    // given
+    val (queue, dlq) = createQueuesAndPopulateDlq()
+
+    // when: start message move task
+    testClient.startMessageMoveTask(DlqArn, maxNumberOfMessagesPerSecond = Some(1))
+
+    // then: ensure that not all messages were moved back to the original queue after 2 seconds
+    Thread.sleep(2000)
+    fetchApproximateNumberOfMessages(queue) should (be > 1 and be < 6)
```
mostr, line 47: Shouldn't that be max 2 and the former one min 4?

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/MessageMoveTaskTest.scala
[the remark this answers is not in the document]
mostr: I've got  the same concern on `Thread.sleep()` making tests brittle and slow but briefly looking didn't find a way to avoid that

mostr (approved):
Few minor comments as I'm quite new to this code

rest/rest-sqs-testing-amazon-java-sdk/src/test/scala/org/elasticmq/rest/sqs/MessageMoveTaskTest.scala
micossow: Only if the message move jobs starts immediately, which I'm not sure is guaranteed in the environments like the Github Actions provide (fractions of CPU core)

Issue #890, opened 2023-10-26
Fifo delay queue is delivering messages immediately instead of delaying them


jcelmeta14:
Hello!

I have a queue configured in the `.conf` file as follows:

```
  MyQueue {
    fifo = true
    delay = 60 seconds
    defaultVisibilityTimeout = 60 seconds
    deadLettersQueue {
      name = "MyDLQueue"
      maxReceiveCount = 3
    }
  }
```

As you may see, the queue has a delay of 60 seconds. The queue seems to have been created correctly. I checked using the CLI:

<img width="606" alt="image" src="https://github.com/softwaremill/elasticmq/assets/22294784/11762444-6d65-4fa3-b77f-5031d9b2b791">

The problem is that regardless of the configuration, the messages get delivered immediately. I am using `serverless-framework` and I have configured a function to have the queue as an event source. As soon as a message is sent to the queue, the function get triggered immediately without the delay of 60 seconds.

Am I doing something wrong?

Thank you!

micossow:
@jcelmeta14 I can't reproduce the issue.
I'm using `softwaremill/elasticmq-native:1.4.7` and the following configuration:
```
include classpath("application.conf")

queues {
  main {
    defaultVisibilityTimeout = 10 seconds
    delay = 2 seconds
    receiveMessageWait = 0 seconds
    deadLettersQueue {
      name = "retry"
      maxReceiveCount = 1
    }
  }
  retry {
    defaultVisibilityTimeout = 10 seconds
    delay = 2 seconds
    receiveMessageWait = 0 seconds
  }
  delayed {
    fifo = true
    delay = 10 seconds
    defaultVisibilityTimeout = 10 seconds
    deadLettersQueue {
      name = "retry"
      maxReceiveCount = 3
    }
  }
}
```

and this is my test code:
```python
queue = self.sqs.get_queue_by_name(QueueName='delayed.fifo')
queue.send_message(MessageBody='some test message to be delayed', MessageGroupId='group123', MessageDeduplicationId='dedup123')
messages = queue.receive_messages()
print(messages)
time.sleep(5)
messages = queue.receive_messages()
print(messages)
time.sleep(5)
messages = queue.receive_messages()
print(messages)
```

and this is the output:
```
[]
[]
[...]

PR #1009, opened 2024-06-05
fix: FIFO queue zero delay

tmnsur:
This PR aligns fifo queue delay behavior with aws sqs when the `DelaySeconds` attribute is sent as 0 in the message
fixes: #890


micossow:
Thanks for the contribution!

PR #1086, opened 2025-01-03
Pass sourceQueueName via System Message Attribute for DeadLetterQueueSourceArn

sergij:
`DeadLetterQueueSourceArn` is part of AWS SQS message metadata that conveys information about the original queue the message was sent to from where it got moved to DLQ. This enables SQS customers to move messages back to the original queue without using `redrive` operation. Original SQS behaviour exposes `DeadLetterQueueSourceArn` on messages received from DLQ, with this change messages received from DLQ from elasticmq will follow the same pattern.


micossow:
Thanks for contribution! I changed it a bit to use a dedicated field in the internal message structure instead of system attributes and added missing changes.

PR #1206, opened 2026-04-05
Add auto-create-queues feature

### LINKED ISSUE #1132: Feature Request: Create queues automaticlly on any http request with no static config
Automatically create queues on any [GET], [POST] HTTP requests and omit static configuration. 

It would be great to have a flag to enable creation of queues on http request - instead of returning not found.

This can eliminate static config files, especially helpful during dev/test cycle, where we don't want to maintain multiple queue configurations, we just want it to do what a mock server needs to do automatically.

I don't mind taking a stab at it... Could someone please suggest which files to look at to start with?
Also, I don't have a Java env setup so im not sure how I'd build and test.

micossow:
## Summary

- Adds a new `auto-create-queues` configuration section. When `enabled = true`, any request targeting a non-existing queue creates the queue on demand instead of returning `NonExistentQueue`.
- An optional `template` sub-section lets users specify default attributes (visibility timeout, delay, DLQ, tags, etc.) for all auto-created queues.
- Queues whose names end in `.fifo` are always created as FIFO queues regardless of the template setting.
- The feature defaults to `enabled = false`, so it is fully backwards-compatible.

Resolves #1132

## Configuration example

```hocon
auto-create-queues {
  enabled = true
  template {
    defaultVisibilityTimeout = 30 seconds
    delay = 0 seconds
    tags {
      env = "dev"
    }
  }
}
```

## Test plan

- [x] Unit tests added to `QueueDirectivesTest` (auto-create disabled returns error, enabled creates queue, `.fifo` detection)
- [x] Integration tests added in `AutoCreateQueuesTests` (send to non-existing queue, FIFO auto-creation, template attributes applied)
- [x] Existing test suite (`restSqs/test`) passes unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ project/boot
 /bin/
 
 .bsp
-.db
+*.db
 .env
 .data
 
--- a/README.md
+++ b/README.md
@@ -164,6 +164,50 @@ queues {
 ```
 If not then suffix will be added automatically during queue creation.
 
+# Auto-creating queues on demand
+
+ElasticMQ can automatically create a queue the first time any request targets it, instead of returning a
+`NonExistentQueue` error. This is useful during development and testing — it removes the need to maintain a static list
+of queue names in the configuration.
+
+To enable, add the following to your configuration file:
+
+```
+auto-create-queues {
+  enabled = true
+}
+```
+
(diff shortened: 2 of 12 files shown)


p10tyr:
This is great. Thank you . Testing it out now!

